PR #106, opened 2026-08-10
fix(debug): stale checkout tokens, paused-permit starvation, silent queue

Bnjoroge1:
## Summary

Fixes the debug-session failure loop that froze the runner pool and left `preloop run` hanging on Queued forever with no shell:

1. **Expired snapshot checkout token.** The checkout token is pinned onto the step at submission and lives ~50 minutes (`LOCAL_JWT_LIFETIME`). Jobs queued past that — or paused in a session that long — failed checkout with a git 401 on claim (`could not read Username … terminal prompts disabled`).
   - the snapshot surface now 401s with a `WWW-Authenticate: Bearer` challenge so git reports the rejection instead of prompting
   - the broker re-mints pinned step tokens **at claim** (steps matched by id, recorded on the message as `preloopSnapshotTokenSteps`)
   - **retry verdicts** carry a fresh credential; the worker swaps it into the replayed steps before `:retry` / `:sync + retry` re-runs them

2. **Paused jobs pinned pool permits.** A debug-session pause blocks the guest worker on a verdict while the host slot task waits for exit, so the slot's concurrency permit stayed held — two unanswered pauses took `max_concurrent` to zero and every later run queued forever.
   - the worker writes a guest pause marker (`/var/lib/preloop-runner/.preloop-job-paused`) while a session is open
   - the pool watches it, releases the slot's permit for the pause's duration, and re-acquires on resume

3. **Silent starvation + clean exits.** `preloop run` now reports queued-job and paused-session counts once a run stalls; detach/abort from the debug prompt exit cleanly instead of printing a bogus `✗ Run: Queued` error.

Also centralizes the worker-debug route paths shared between the router and the control-socket allowlist (`auth.rs`), so a route edit can no longer silently drop worker-bearer coverage.

## Verification
- New regression tests: claim-time re-mint, verdict-carried token, worker token refresh (pinned-only), snapshot Bearer challenge, pool permit release/re-acquire, redirect step-id recording
- `cargo fmt --all --check`, `cargo clippy --workspace --all-targets`, full workspace tests green
- Live repro: failing step under a TTY opens the `[d/r/s/a]` shell immediately

---
## Summary by cubic
Fixes a debug-session feedback loop that froze the runner pool and left runs queued without a shell. Snapshot checkout tokens are refreshed, paused jobs no longer pin permits (with correct provider exit semantics), and `preloop run` now surfaces real backpressure and explains when failure shells are disabled.

- **Bug Fixes**
  - Snapshot checkout: server 401s with a `WWW-Authenticate: Bearer` challenge; broker re-mints pinned step tokens at claim; retry verdicts carry a fresh token; the worker swaps it only into steps marked by `preloopSnapshotTokenSteps`.
  - Paused sessions: the worker writes a pause marker; the pool releases the slot’s permit while paused and re-acquires it on resume; absent-marker probes (exit 0/1) are handled as real results; transport failures preserve the last known state with per-machine, rate-limited warnings; slow re-acquires are logged.
  - Run UX: `preloop run` reports queued-job and paused-session counts only when capacity is actually blocking; the failure prompt opens only for `SessionState::Paused`; warns once if it can’t poll sessions; shows a one-time notice when failure shells are off (detached run or non-TTY), suppressed under `CI`.

- **Refactors**
  - Centralized worker-debug route paths and member suffixes shared by the router and control-socket allowlist to keep bearer coverage in sync.

<sup>Written for commit dad9762accd382d426c1b2bc7351bc9dba81433f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/preloopdev/preloop/pull/106?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/preloop-cli/src/debug_session.rs
+++ b/crates/preloop-cli/src/debug_session.rs
@@ -163,23 +163,18 @@ struct Api {
     token: Option<String>,
 }
 
-/// Find a paused session belonging to a run, if one exists.
-pub async fn paused_for_run(
+/// All open sessions on the server, newest first.
+pub async fn list_sessions(
     client: &reqwest::Client,
     base_url: &str,
     token: Option<String>,
-    run_id: preloop_gha_protocol::RunId,
-) -> Option<DebugSession> {
+) -> Result<Vec<DebugSession>> {
     let api = Api {
         client: client.clone(),
         base_url: base_url.to_owned(),
         token,
     };
-    api.list()
-        .await
-        .ok()?
-        .into_iter()
-        .find(|session| session.run_id == run_id)
+    api.list().await
 }
 
 /// Offer the choice at the moment of failure, inline in `preloop run`.
--- a/crates/preloop-cli/src/main.rs
+++ b/crates/preloop-cli/src/main.rs
@@ -1365,7 +1365,9 @@ async fn cmd_run(args: RunArgs) -> anyhow::Result<()> {
     // waiting for a controller), but a user who expects the debug shell on
     // failure needs to know it was disabled before the run fails — the
     // alternative is a plain `✗` with no explanation and no way to attach.
-    if !submission.preserve_on_failure && !args.no_debug {
+    // A piped pipeline never expects a failure shell, so the notice would be
+    // noise on every CI run; it targets interactive-ish invocations only.
+    if !submission.preserve_on_failure && !args.no_debug && std::env::var("CI").is_err() {
         eprintln!(
             "[preloop] failure shells are off ({}); pass --preserve-on-failure to \
              pause on failure for `preloop debug`/`preloop shell`",
@@ -1495,6 +1497,8 @@ async fn cmd_run(args: RunArgs) -> anyhow::Result<()> {
                             if let Some(session) =
                                 sessions.into_iter().find(|session| {
                                     session.run_id == accepted.run_id
+                                        && session.state
+                                            == preloop_gha_protocol::debug_session::SessionState::Paused
                                 })
                             {
                                 paused = Some(session);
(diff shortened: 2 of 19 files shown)


Bnjoroge1:
All nine findings verified against the code and addressed (commit `784f6d1b`):

**Fixed:**
- **orchestrator: probe failure ≠ resumed** — a failed guest probe (transport/smolvm error) now preserves the last known pause state with a warn, instead of being read as "resumed" and re-pinning the permit mid-pause.
- **orchestrator: over-subscription transparency** — the resume path keeps the blocking permit re-acquire (so future forks stay bounded by `max_concurrent` + paused) and now logs when the acquire is delayed, surfacing the transient window. A hard gate needs a host/worker resume handshake (worker waits for host capacity before continuing) — that's a protocol change I've flagged in the code comment as the follow-up; the current window is bounded by the 2s poll interval.
- **cli: notice noise** — the "failure shells are off" notice is now suppressed under `CI` env (pipelines never expected a shell).
- **cli: non-paused sessions** — only `SessionState::Paused` sessions trigger the failure prompt; a resolved/closed session no longer interrupts reporting.
- **cli: backpressure false positive** — the stall line only fires when jobs or sessions are actually waiting; a slow in-progress step is not a stall.
- **worker: marker I/O** — pause-marker write/remove failures are `warn!`ed instead of `let _ =` (the permit-release mechanism can no longer fail silently).
- **routes: suffix drift** — member routes now build from the shared suffix constants, so the allowlist guard covers the suffixes too.
- **broker: allocation-free re-mint** — pinned ids are parsed to a `HashSet<Uuid>` once; no per-step stringification.
- **tests: real claim path** — new integration test submits a local-workspace workflow, claims it through the broker, and asserts the acquired checkout carries a freshly minted token (deterministic via a 1.1s clock advance).

Full suites green: server 485, runner 519, orchestrator 43+14, cli 113; fmt clean.

Bnjoroge1:
Follow-up review addressed (commit `dad9762a`) — and the P1 was a real bug, thanks:

1. **P1: resume never re-acquires with the real provider** — confirmed: `SmolVmProvider::checked_with_network` maps *any* nonzero smolvm exit (including the guest `test -f` exit 1) to `Err(VmError::Command)`, so my state-preserving error path swallowed the normal absent-marker result and the permit was never re-acquired after a resume. The watcher now treats `VmError::Command` with exit 0/1 as a real probe result (`test -f` only ever exits 0/1); only genuine transport failures (other variants) preserve state. Regression coverage: the probe provider now models the real provider's `Err(VmError::Command{exit:1})` for an absent marker, and a new test (`probe_transport_errors_preserve_pause_state`) proves transport errors don't re-pin a paused job's permit.

2. **P3: probe-warning spam** — the per-machine warning is now rate-limited to once per 60s while the probe keeps failing.

3. **P3: warn in Drop** — `NotFound` is no longer warned (the marker may legitimately never have been written or already be gone); Drop stays quiet on the normal path and only reports genuine removal failures.

Full suites green: orchestrator 44+14, runner 519, server 485, cli 113; fmt clean.

PR #112, opened 2026-08-10
fix(pool): serialize forks per golden, re-arm a spent fork base, resolve debug refs by run

Bnjoroge1:
## Problem

A run stalled with `2 job(s) queued, 2 debug session(s) paused` and no debug shell anywhere: `preloop debug <run-id>` answered `no paused job matching`, and the queue never moved. The engine log showed the underlying failure:

```
smolvm fork failed with exit code 1: Freezing golden 'preloop-runner-golden' as fork base...
Error: agent operation failed: fork: golden 'preloop-runner-golden' is already paused; a valid retained checkpoint is required
```

### Root cause

SmolVM keeps **one RAM checkpoint per golden** (its own fork-pool controller serializes on the golden for exactly this reason: "a golden can produce only one RAM checkpoint at a time"). Preloop's on-demand pool forks the same golden **concurrently** — one `machine fork` per queued job, unserialized. The concurrent forks race the freeze/checkpoint; the loser's rollback **resumes the golden and deletes the checkpoint**. Since the plain `machine fork` path preloop uses never persists a reusable checkpoint (`prepare_forks_reusing(..., None, false)`), a paused golden with no valid checkpoint fails **every** subsequent fork identically. The fallback (direct VM create from the packed artifact) is slow, and nothing ever re-arms the golden — the queue stalls until the engine restarts and rebuilds it.

On a dev machine this reproduced live: 20,405 checkpoint directories (~800 MB) accumulated under the golden's data dir, the fingerprint of repeated fork races (each rollback resumed the golden, forcing the next fork to re-checkpoint).

### Why the debug terminal never came up

The two paused sessions belonged to an *earlier* run. The stalled run's jobs never got a runner, so there was nothing to attach — and `preloop debug <run-id>` 404'd because the run id matched no session, with the error message not saying what *was* paused.

## Changes

- **`preloop-vm`** — serialize `machine fork` per golden: one in-flight fork per golden, keyed by name; different goldens still fork in parallel. Track forked clones (name → golden) so a spent base can be re-armed safely; `delete` clears the census. `VmProvider::has_live_forks` defaults to `true` (conservative: refuse re-arm) for providers that don't track clones.
- **`preloop-orchestrator`** — when a fork fails with the spent-base signature and **no clone depends on the golden**, stop → `start --forkable` → retry the fork once: the queue recovers in seconds instead of stalling until an engine restart. A base with live clones is never restarted (resuming it would corrupt the copy-on-write clones); it falls back to direct creation with an error-level log explaining why.
- **`preloop-cli`** — `preloop debug <reference>` now resolves a run id that has several paused jobs: it lists them (session id, job, run, step) instead of answering `no paused job matching`. When the reference matches nothing but other sessions are paused, the error lists what is paused. The no-argument listing now shows each session's run id.
- CHANGELOG entry under Unreleased.

## Verification

- New unit tests:
  - `preloop-vm`: `forks_from_one_golden_never_overlap` (barrier-proven), `forks_from_different_goldens_still_overlap`.
  - `preloop-orchestrator`: spent base with no live clones → re-arm sequence (fork, cleanup, stop, start-forkable, fork retry); spent base with live clones → no re-arm, fallback + cleanup; existing fallback test still green.
  - `preloop-cli`: run id with two paused jobs resolves to both; session id resolves exactly; unrelated run id matches nothing; session line includes run id.
- Full suites: `preloop-vm` 17 ✓, `preloop-cli` 119 ✓, `preloop-orchestrator` 78 ✓.
- End-to-end smoke on a Mac: two-job workflow ran green twice (cold 148s / warm 38s) with 11 serialized forks from one golden and zero failures; `preloop debug <run-id>` demonstrated resolving a live paused session; leftover sessions aborted cleanly.

## Follow-up (not in this PR)

- Replacing a spent golden with a fresh generation (rather than re-arming the same base) so recovery works even while paused debug VMs hold live clones.
- smolvm-side: garbage-collect accumulated checkpoint directories per golden (`s/` grows one small dir per fork).

---
## Summary by cubic
Fixes stalled queues by serializing forks per golden and auto-rearming a spent fork base when safe. Also improves `preloop debug` to resolve paused sessions by run so users can attach reliably.

- **Bug Fixes**
  - `preloop-vm`: serialize `machine fork` per golden; different goldens still fork in parallel. Track forked clones and clear the census on `delete`; deleting a golden evicts its clones from the census. Add `rearm_fork_base` in the provider, which runs under the same per-golden lock as `fork`: it cleans up a partial clone (or accepts “not found”), verifies no live clones, then `stop` → `start --forkable`; default providers return `false` to fall back safely.
  - `preloop-orchestrator`: on a spent-base fork failure with no live clones, call provider `rearm_fork_base` and retry once; operations are atomic with forking. If live clones exist or cleanup fails, do not re-arm; fall back to direct create with a clear error.
  - `preloop-cli`: `preloop debug <reference>` resolves run IDs with multiple paused jobs and lists all matches. The “these are paused instead” hint only replaces a 404; transport/5xx errors bubble up unchanged.

<sup>Written for commit a6ca1b1248865a8cc188c79eeb84b177be4da8db. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/preloopdev/preloop/pull/112?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,24 @@ Releases before v0.27.0 predate the changelog.
 
 ## [Unreleased]
 
-## [0.29.9] - 2026-08-10
+### Fixed
+
+- Serialize `machine fork` per golden VM. SmolVM keeps one RAM checkpoint per
+  golden; concurrent forks raced the freeze, and the loser's rollback resumed
+  the base and dropped the checkpoint, after which every fork failed with
+  `golden '<name>' is already paused; a valid retained checkpoint is required`
+  and queued jobs stalled until the engine restarted. Forks from different
+  goldens still run in parallel.
+- Re-arm a spent fork base (stop + restart forkable) and retry the fork once
+  when no clone still depends on it, instead of degrading every runner to a
+  full VM create. A base with live clones is never restarted — it falls back
+  to direct creation with an error explaining why.
+- `preloop debug <reference>` now resolves a run id that has several paused
+  jobs: it lists them (with their run ids) instead of answering
+  `no paused job matching`. When nothing matches but other sessions are
+  paused, the reply says what is paused instead of a bare 404.
+
+## [0.29.8] - 2026-08-09
 
 ### Added
 
--- a/crates/preloop-cli/src/debug_session.rs
+++ b/crates/preloop-cli/src/debug_session.rs
@@ -82,7 +82,7 @@ pub async fn run(
 
     let sessions = ctx.list().await?;
     let session = match (&args.session, sessions.len()) {
-        (Some(reference), _) => ctx.get(reference).await?,
+        (Some(reference), _) => resolve(&ctx, &sessions, reference).await?,
         (None, 0) => {
             println!("No paused jobs.");
             println!();
@@ -97,14 +97,7 @@ pub async fn run(
             println!("{} paused jobs:", sessions.len());
             println!();
             for session in &sessions {
-                println!(
-                    "  {}  {}  step {}/{} {}",
-                    session.session_id,
-                    session.job_name,
-                    session.step.index + 1,
-                    session.step.total,
-                    session.step.display_name
-                );
+                println!("{}", session_line(session));
             }
             println!();
             println!("Attach with: preloop debug <session-id>");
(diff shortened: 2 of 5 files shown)


Bnjoroge1:
All review findings addressed in a6ca1b12:

- **P0 (atomicity):** re-arm moved into the provider as `rearm_fork_base`, taking the same per-golden lock as `fork` — partial-clone cleanup, the live-clone check, and stop/start-forkable are atomic with new forks. A concurrent slot can no longer create a clone between the empty check and the golden being stopped.
- **P0 (cleanup gating):** re-arm only proceeds when cleanup positively succeeds ("not found" counts as positively gone); a failed delete blocks the re-arm and falls back.
- **P2 (census):** deleting a golden now evicts every clone forked from it, so a rebaked golden name no longer reports live clones forever.
- **P3 (CLI):** the "these are paused instead" listing replaces only the 404 case; transport/5xx errors propagate their real cause.

Tests: orchestrator 53, vm 17, cli 24 — all passing, including the new failed-cleanup and re-arm-sequence cases.

PR #113, opened 2026-08-11
fix(github): persist check-run ids at creation so restarts cannot orphan checks

Bnjoroge1:
## Problem

GitHub checks stayed `queued` forever while the jobs actually ran and completed. On the dogfood server: webhook burst at 22:38 UTC created check runs for 5 runs; a service restart at ~23:02 (release update) happened while the jobs were still queued behind long-running jobs; one of those jobs (`rust`) completed Succeeded at 00:03:51 — and GitHub's check for it still says `queued`.

### Root cause

`job_check_run_ids` lived only in server memory. Run records are persisted to SQLite only on job-status events, and a queued job emits none until it starts. A restart while the queue is backed up restored the runs with an **empty mapping**, and every later `in_progress`/`completed` check update silently returned at `None => return` — no log, no retry. Any deploy during a busy queue orphans every check created since the previous status event.

## Changes

- **`preloop-gha-protocol`**: new `NdjsonEvent::CheckRunCreated` — a persistence trigger, not a timeline event.
- **`preloop-runner-server`**: `report_check_run_queued` (and the rerequest-reuse path in `rerun_run_inner`) now emit `CheckRunCreated` the moment a check id is recorded, persisting the run projection immediately instead of on the next job event.
- **Test**: `check_run_ids_survive_a_restart_before_any_job_event` — webhook → mock check id → full store round-trip with no job event in between → mapping intact. **Verified failing without the persist** (temporarily reverted, test failed) and passing with it.

Also on this branch (unreleased): `fix(server): requeue dead-bound jobs at the back of the pool waitlist` — jobs whose machine died before claiming no longer re-pair with unconditional priority; stale pool-pending marks stay visible to dispatch (observed in production: 9 re-pairs in 48 min starving newer jobs).

## Verification

- `cargo test -p preloop-runner-server --lib`: 493 passed, 0 failed (new test included).
- `cargo fmt --check` and clippy clean on the touched crates.
- Branch rebased onto `v0.29.9`.

## Recovery for already-orphaned checks (no deploy needed)

Click **Re-run** on the stuck checks in GitHub: the `check_run` rerequest webhook re-links the check run id (falling back to matching by job name when the stored mapping is empty) and resubmits the run, so the new attempt reports `in_progress`/`completed` normally.

---
## Summary by cubic
Prevents GitHub checks from staying queued by persisting check-run IDs at creation. Also fixes scheduling so stale bindings can’t starve newer jobs, and keeps stale bindings claimable in strict non‑pool mode.

- **Bug Fixes**
  - `preloop-gha-protocol`: add `NdjsonEvent::CheckRunCreated` to persist the run as soon as check-run IDs are known; `preloop-runner-server` emits it from `report_check_run_queued` and the reuse path in `rerun_run_inner`.
  - Test: check-run IDs survive a restart before any job event.
  - Scheduler: release stale job→runner bindings, requeue them at the back of the waitlist, preserve `first_at`, and still offer stale pool-pending marks; strict non-pool mode keeps stale bindings claimable. Adds regression tests.
  - Recovery for already-orphaned checks: click Re-run in GitHub to re-link and report normally.

<sup>Written for commit 1f34924e41e14b56dba30e46b47a7cc74412092b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/preloopdev/preloop/pull/113?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Bnjoroge1:
All review findings addressed in 1f34924e:

- **P1 (strict non-pool stranding):** stale bindings are no longer cleared when there is no pool waitlist to re-mark on — the record survives and strict `claim_permitted` keeps letting verified runners take it over. New test `strict_non_pool_mode_keeps_a_stale_binding_claimable` (verified failing without the fix).
- **P1 (first_at reset):** the requeue preserves the original first-bound stamp (record kept, `at` re-stamped) so the replacement pairing cannot extend the bounded claim window; assertions added to the requeue test.
- **P2 (test env):** the check-run restart test restores caller-provided GitHub env vars via an RAII guard.
- **P2 (docs):** direct snapshot example sets `PRELOOP_RUNNER_STORAGE_GB=80`.

Full server suite: 494 passed.

PR #119, opened 2026-08-11
Strengthen golden VM image provenance

Bnjoroge1:
## Summary

- Capture digest-verified OCI index, selected platform manifest, attached attestation descriptors, and upstream SPDX SBOM for each Linux golden base.
- Fail closed on mutable base references for release golden builds and bind the exact base evidence to a deterministic provenance manifest.
- Sign and GitHub-attest the golden plus its base-evidence, SBOM, and provenance sidecars; preserve and publish those sidecars when reusing prior goldens.
- Document the mirror.gcr.io provenance model, verification commands, and the new digest policy.

## Validation

- `cargo test -p preloop-cli` — 142 passed.
- `just test-ci` — passed fmt-check, clippy, workspace tests, and conformance.
- `PRELOOP_PORT=63990 just dogfood` — passed; `job_succeeded=1`, `e2e_latency_ms=350815` (the default port was already occupied by an existing local server).
- Direct amd64/arm64 base-evidence capture, SPDX extraction, provenance-manifest smoke test, Python compilation, and release-workflow YAML parsing passed.

No live release publication was performed.

---
## Summary by cubic
Hardens golden VM image provenance by capturing exact base image evidence and SBOM, digest‑pinning release bases, and signing/attesting all artifacts. Previously we allowed mutable tags and lacked base evidence; now we bind the golden to the exact base index/platform digest and publish verifiable sidecars without changing build outputs.

- Adds `scripts/capture-base-provenance.py` to record the digest‑verified index, selected platform manifest, and attached attestations; discovers via the OCI referrers API, unwraps in‑toto SPDX predicates to a bare SPDX JSON sidecar, and enforces a source‑prefix match on a path boundary.
- Adds `scripts/write-golden-provenance.py` to emit a deterministic `<golden>.provenance.json` that ties the golden hash to the base index/platform digest and workflow; verifies the base SBOM’s digest against captured evidence before binding.
- Release workflow resolves the golden identity/base where consumed, captures base evidence and SBOM, attests the golden plus sidecars in one GitHub attestation, and uploads/preserves sidecars (`.base-provenance.json`, `.base-sbom.spdx.json`, `.provenance.json`) including when reusing a prior golden.
- Enforces digest‑pinned bases for releases via `PRELOOP_REQUIRE_BASE_DIGEST=1`; local file inputs are exempt. Docs add verification commands and clarify the `mirror.gcr.io` cache model; CLI docs add `PRELOOP_VERIFY_BASE_IMAGE` and `PRELOOP_REQUIRE_BASE_DIGEST`.

- Migration
  - Use digest‑pinned base references (`...@sha256:<digest>`). Set `PRELOOP_REQUIRE_BASE_DIGEST=1` to fail local builds on mutable tags; local file bases continue to work.
  - To verify a release, run `cosign verify-blob` on the `.bundle` and `gh attestation verify` on the golden.

<sup>Written for commit 9243fbee829aea6f9384fc0afb885597133a1ad8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/preloopdev/preloop/pull/119?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/release-golden.yml
+++ b/.github/workflows/release-golden.yml
@@ -43,11 +43,10 @@ jobs:
             --pattern 'preloop-ubuntu-24.04-*' \
             --dir dist \
             --repo "$GITHUB_REPOSITORY"
+          # Preserve the signatures, attestations, and base-evidence sidecars
+          # when a release intentionally reuses an earlier golden.
           gh release upload "$TARGET_TAG" \
-            dist/preloop-ubuntu-24.04-aarch64 \
-            dist/preloop-ubuntu-24.04-aarch64.sha256 \
-            dist/preloop-ubuntu-24.04-x86_64 \
-            dist/preloop-ubuntu-24.04-x86_64.sha256 \
+            dist/preloop-ubuntu-24.04-* \
             --clobber \
             --repo "$GITHUB_REPOSITORY"
 
@@ -70,6 +69,16 @@ jobs:
     steps:
       - name: Check out repository
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+
+      - name: Resolve golden identity and base
+        run: |
+          set -euo pipefail
+          echo "RUNNER_TARGET=aarch64-unknown-linux-gnu" >> "$GITHUB_ENV"
+          echo "GOLDEN_ARCH=aarch64" >> "$GITHUB_ENV"
+          echo "GOLDEN_PLATFORM=linux/arm64" >> "$GITHUB_ENV"
+          BASE_IMAGE="$(awk -F'"' '/^ubuntu_24_04_base = / { print $2; exit }' versions.toml)"
+          test -n "$BASE_IMAGE"
+          echo "BASE_IMAGE=$BASE_IMAGE" >> "$GITHUB_ENV"
 
       - name: Set up Rust toolchain
         uses: dtolnay/rust-toolchain@4360b52568e2003a75bf9bc1d59f33a8e3fc893c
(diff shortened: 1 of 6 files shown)


Bnjoroge1:
Addressed all review findings in c6041a7e:

- **P0** (cubic + codex, release-golden.yml): removed the dead `Resolve golden identity and base` step from `build-and-publish-linux-binaries`; added it to `build-and-publish-golden` (Linux X64) right after checkout with the correct x86_64/linux-amd64 values. The macOS golden job keeps its aarch64 copy.
- **P1** (cubic, capture-base-provenance.py): source-annotation matching now requires an exact prefix match or a `/` path boundary — a lookalike URL sharing the prefix string is rejected.
- **P2** (cubic, write-golden-provenance.py): the base SBOM sha256 is verified against the base-evidence digest, and a missing SBOM record fails the write instead of binding an unrelated SBOM.
- **P2** (codex, capture-base-provenance.py): SPDX attestations stored as in-toto statements (BuildKit format) have their `predicate` unwrapped so the published `.base-sbom.spdx.json` is a bare SPDX JSON document; the evidence records both the published sidecar digest and the original registry blob digest.
- **P3** (cubic, main.rs): `require_digest_pinned_base` exempts existing local files (`Path::exists()`), so relative packed-artifact paths like `dist/preloop-ubuntu-24.04.smolmachine` work with `PRELOOP_REQUIRE_BASE_DIGEST=1`.

Additional fix while wiring the SBOM binding: attestations are discovered via the OCI referrers API with an index-manifests fallback — cosign-attached attestations do not rewrite the subject index, so index-only discovery missed them.

`just test-ci` passes locally (fmt, clippy, full workspace tests, runner-watch conformance).

PR #122, opened 2026-08-11
fix(runner): drop-in workflow fidelity, golden bake, pool/check-run hardening

Bnjoroge1:
## Summary

Two threads of work on the branch:

### Pool / check-run reliability (earlier commits)
- `fix(server): requeue dead-bound jobs at the back of the pool waitlist` — f83ab197
- `fix(github): persist check-run ids at creation so restarts cannot orphan checks` — 25076723
- `fix(scheduler): keep stale bindings claimable in strict non-pool mode; preserve first_at` — 1f34924e

### Drop-in workflow fidelity (56b15324)

Running a real workflow (runner-image-blobs' attest pipeline) against the
local runner surfaced four fidelity gaps where unmodified GitHub workflows
break on preloop:

1. **Per-step `shell:` dropped from job messages** — `TaskStep` never emitted
   `inputs.shell`, so `shell: mv {0} script.js` steps ran under bash and
   script content was parsed by the wrong interpreter. The official
   ScriptHandler reads `Inputs["shell"]`; the wire message now carries it.
2. **Step-level `env:` missing from the `env` context** during the step's own
   script evaluation — `run: ${{ env.CMD }}` with step-level `env: CMD:`
   resolved empty. `execute_step` now overlays the step env onto the
   expression context (`Context::merge_root`).
3. **Type-3 format expression tokens collapsed to empty** — values containing
   `${{ }}` serialize as `format('...', ...)` tokens; the runner only read
   `lit`, so scripts/env collapsed to `''`. Extraction now keeps the raw
   `expr`, and `evaluate_template` evaluates bare format expressions.
4. **Golden bake ownership + parity** — image layers are extracted on the
   host (cannot chown to root), so `/usr/bin/sudo` lost root ownership and
   its setuid bit and `/etc/sudoers` checks failed. The bake now re-owns the
   system paths, restores setuid bits, and installs the gh CLI (hosted
   parity).
5. **Packed-golden artifact invalidation** — the pack was keyed on the base
   image only, so bake-content changes never rebuilt the golden (the
   documented "fingerprint-suffixed artifact" key was never implemented).
   The artifact payload now includes the environment fingerprint.

Also fixes a pre-existing clippy `needless_borrow` and allows a
pre-existing `too_many_arguments` on `provision_runner`.

## Validation
- `just test-ci` fully green: fmt, clippy (workspace, all targets), all
  tests, and `conform` against the v2.336.0 golden wire captures.
- End-to-end on a real runner: custom-shell steps, step-env expressions,
  and the attest workflow's checkout/discovery/artifact-download steps
  execute correctly in a golden-forked VM.


Bnjoroge1:
Push back: this is not merge-ready; the current rust check is failing and the latest bot review still has P1s.

The highest-risk unresolved items are:
- Persisted/reloaded `TaskStep` data loses type-3 expression provenance, so restart can turn expressions into literals and change scripts, inputs, or shell behavior.
- The template fix infers expression provenance from text (`format('...')`), which executes legitimate literal values, strips leading whitespace via `trim_start()`, and still mishandles bare type-3 format values containing a literal `${{`.
- The golden-bootstrap ownership/mode change has no failing reproduction or guest filesystem regression test; validating only the generated shell is insufficient for this sensitive path.
- The explicit `build-golden --output` path is now fingerprint-suffixed while the CLI still checks the unsuffixed path, so a successful build can be reported as failed.
- The README CLI-reference link and several VM-image instructions are broken regressions.
- Check-run persistence and stale-job replacement need bounded/deterministic behavior rather than a full projection write per matrix job and timestamp ties.

Please fix the P1s and add the guest/protocol regression coverage before merge. The currently failing rust check must be green afterward.

PR #138, opened 2026-08-16
fix(security): cap matrix expansion at 256 combinations

Bnjoroge1:
## Summary

`strategy.matrix` cross-products were fully materialized with no ceiling. A ~450 byte workflow (10 axes × 10 values = 10¹⁰ combinations) drove `Vec::with_capacity` into a saturating huge allocation and the expansion loop into unbounded memory growth — one authenticated `POST /api/v1/runs` kills the server.

**Reproduced** (release build, pre-fix): process killed after 71s at 1.85 GB resident / 29.8 GB virtual. Post-fix the same payload is rejected cleanly:

```
REJECTED CLEANLY: matrix for job `bomb` can not generate more than 256 jobs
```

## Changes

- `try_expand_matrix_spec`: refuses specs whose cross-product **or final row count** (include-only rows can append past the cross product) exceeds 256, GitHub's documented per-job matrix limit.
- All three production expansion sites (static matrix, expression matrix, deferred `needs`-dependent matrix) now route through it.
- New `ParserError::MatrixTooLarge` mirrors GitHub's own error message, so wire behavior matches what the official runner sees upstream.

## Tests

- cross-product over limit → `MatrixTooLarge` before materialization
- include-only overflow → `MatrixTooLarge`
- exactly 256 → still expands
- end-to-end `parse_workflow` + `expand_jobs` with the bomb YAML → `MatrixTooLarge`

`just test-ci`: green except the pre-existing `test_golden_acquirejob_payloads_parsing` failure, which also fails on `main` (missing `06-multi-step` golden locally).

---
## Summary by cubic
Caps `strategy.matrix` fan-out at 256 jobs and bounds expansion work to prevent OOM/DoS while matching GitHub’s limit. Previously we materialized the full cross‑product with no ceiling; now oversize matrices fail fast with `ParserError::MatrixTooLarge`. Golden image downloads also gain bounded timeouts, progress, and clearer errors while still falling back to local builds.

- All expansion sites (static, expression, deferred `needs`) route through `try_expand_matrix_spec`; excludes apply before counting; include-only rows count toward the same 256 cap; exactly 256 still expands. Expansion work is capped (~1e6 steps) and axes are bounded; added `MatrixTooLarge` error.
- Removed an undocumented payload-size limit. Golden downloads use 2h timeouts, emit size-aware progress and bytes, and surface OCI/release failures (auth, digest mismatch, truncation). When `Content-Length` is absent, OCI downloads fall back to the layer’s `size`.

**Migration**
- Reduce any workflow that would generate more than 256 jobs from a single matrix.
- Handle `ParserError::MatrixTooLarge` in callers that invoke matrix expansion.

<sup>Written for commit 07a2b545c8a718d8f2a85b0a0f9f4b3a56e51217. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/preloopdev/preloop/pull/138?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Matrix-based workflows now reject expansions exceeding 256 job combinations.
  * Clear validation errors identify the affected job and permitted limit.
  * Matrix expansion errors are now consistently reported during workflow processing.
  * VM and OCI downloads now provide clearer errors and progress reporting.
  * Release downloads can verify SHA-256 checksums before publishing.
  * Registry authentication and HTTP failures are reported instead of being silently ignored.

* **Tests**
  * Added coverage for matrix limits, download progress, checksums, and streaming failures.


Bnjoroge1:
Push back: this security fix is not merge-ready; the latest required rust check is also failing.

The unresolved bot findings are substantive:
- The raw Cartesian-product guard rejects matrices whose `exclude` reduces the effective rows to 256 or fewer (including all-excluded cases), which is stricter than the effective job-count contract.
- Include-only overflow is checked only after `expand_matrix_spec` has fully materialized and preallocated the rows, so a large include list can still drive the targeted OOM before rejection.
- Matching include entries with many distinct non-axis keys are cloned into all 256 rows; a row-count-only cap does not bound total payload/work and leaves an amplification path.

Please make expansion bounded during construction, define whether the 256 limit applies before or after excludes/includes based on official-runner behavior, and add regression coverage for both effective-count and include-amplification cases before merge.

PR #139, opened 2026-08-16
fix(security): cap expression parser nesting depth

Bnjoroge1:
## Summary

The recursive-descent parser (and the evaluator that walks the same AST) had no depth ceiling. An expression of 100k nested parens — a 200 KB payload, far under the 64 MiB run-submission body limit — overflowed the thread stack, which **aborts the process** rather than unwinding. Reachable from any authenticated workflow submitter via `${{ }}` interpolation in run creation (`runs.rs` `eval_expression`), and from repo-controlled YAML via the webhook path.

**Reproduced** (release build, pre-fix): `thread 'main' has overflowed its stack — fatal runtime error: stack overflow, aborting`, exit 134. Post-fix the same payload is rejected cleanly:

```
REJECTED CLEANLY: expression nesting exceeds the maximum depth of 256
```

## Changes

- Depth counter in the two recursive parse functions: paren/argument nesting through `parse_expr`, negation chains through `parse_unary` (`!!!…x` recurses through a different function than parens and needs its own guard).
- New `ExpressionError::TooDeep` past 256 levels. Dotted access chains (`a.b.c…`) and the lexer are iterative and need no guard.

## Tests

- 100k nested parens → `TooDeep` instead of stack overflow
- 100k chained `!` → `TooDeep`
- 200 levels of parens / negations → still evaluate correctly

`just test-ci`: green except the pre-existing `test_golden_acquirejob_payloads_parsing` failure, which also fails on `main` (missing `06-multi-step` golden locally).

---
## Summary by cubic
Caps expression nesting at 256 to prevent stack overflows that aborted the process. Previously, unbounded nesting in `${{ }}` or repo YAML could crash the runner; now inputs over the limit return `ExpressionError::TooDeep` and are rejected cleanly.

- Enforces depth in `parse_expr` (paren/argument nesting), `parse_unary` (negation chains), and binary chains in `parse_or`, `parse_and`, and `parse_eq`; dotted access and the lexer remain iterative.
- Adds `ExpressionError::TooDeep(usize)` with “expression nesting exceeds the maximum depth of 256”.
- Adds boundary tests: accepts max-depth-1 expressions; rejects over-limit parens, long negation chains, and long `||`/`&&`/`==` chains.

<sup>Written for commit d431f86c5c1c8519d1b137bda6ff9efa59f3bc3e. Summary will update on new commits.</sup>

[...]

PR #140, opened 2026-08-16
fix(security): cap format() output at 1 MiB

Bnjoroge1:
## Summary

`format()` multiplies its arguments into the output: each nested `format('{0}{0}{0}', …)` triples its argument, so a **1.4 KB expression** (20 nesting levels around a 1 KB literal) demands ~3 GB of string allocation — far under any body-size limit, since the blow-up is multiplicative per level, not linear in input.

**Reproduced** (release build, pre-fix): process killed by the memory-pressure killer at 6.2 GB resident / 80 GB virtual (exit 137). Post-fix the same payload is rejected cleanly:

```
REJECTED CLEANLY: format() output exceeds the maximum of 1048576 bytes
```

## Changes

- `MAX_FORMAT_OUTPUT_BYTES = 1 MiB` enforced at every append in `format_args` (template segments, argument renders, and escaped braces), via a small `push_capped` helper.
- New `ExpressionError::FormatOutputTooLarge`. 1 MiB is ~6 orders of magnitude above real workflow usage, so no legitimate expression changes behavior.

Note: compose this with #139 (depth cap) and the worst case for nested `format()` is fully bounded.

## Tests

- 20-level amplification bomb → `FormatOutputTooLarge`
- normal formatting unchanged; output just under the cap passes through intact

`just test-ci`: green except the pre-existing `test_golden_acquirejob_payloads_parsing` failure, which also fails on `main` (missing `06-multi-step` golden locally).

---
## Summary by cubic
Caps format() output at 1 MiB and enforces an 8 MiB evaluation budget across expression evaluation to block memory amplification and OOMs. Previously nested format() could allocate gigabytes and templates/conditions preserved the raw token on failure; now oversize output raises FormatOutputTooLarge and over‑budget intermediates raise EvaluationTooLarge, and both propagate from templates and conditions.

- format() applies the cap on every append and when rendering non‑string arguments via JSON serialization bounded by the remaining format capacity (large containers cannot expand during formatting).
- join() respects the evaluation budget and caps intermediate concatenations; the budget counts container/null overhead, and case() remains lazy. The evaluator is refactored to keep recursion stack‑thin under the existing depth cap without changing behavior.

**Migration**
- Handle FormatOutputTooLarge and EvaluationTooLarge from expression evaluation; templates and conditions now surface these errors instead of returning the raw token.

<sup>Written for commit 4d4f3a73cc0c24fb99624d761e989ca0a37e296c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/preloopdev/preloop/pull/140?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Added safeguards against excessively large expression evaluations, joined results, and formatted outputs.
  * Resource-limit errors now correctly propagate through templates and conditions instead of appearing as literal expressions.
  * Preserved existing formatting validation and behavior for valid outputs.

* **Tests**
  * Added coverage for oversized, nested, and near-limit formatting and expression evaluations.
  * Verified consistent resource-limit error reporting during template and condition evaluation.


Bnjoroge1:
Push back: not merge-ready despite green CI.

The latest bot findings still describe correctness/security gaps in the cap itself:
- `join_args` constructs the joined string before charging the evaluation budget, so a repeated separator can allocate far beyond the 8 MiB budget.
- `value_size()` undercounts `Null` and container/element overhead, and evaluation charges after cloning/resolving values; large arrays can therefore evade the budget before rejection.
- The raw format-template length check rejects safe rendered output (for example escaped-brace-heavy templates whose rendered result is under 1 MiB).
- The template regression only asserts `is_err()`, not that `FormatOutputTooLarge` propagated, so a wrong parser/formatting failure can satisfy it.

Please fix the first two before merge and add the boundary/error-variant tests. A green run does not prove the memory guard is effective against the reported amplification paths.

PR #148, opened 2026-08-18
fix(server): cap per-job live log buffer to bound memory

Bnjoroge1:
## Problem

The live console feed ingestion paths — the runner WebSocket (`/ws/live-logs/:job_id`)
and the `TimeLineWebConsoleLog` POST — pushed every batch into an unbounded
per-job `Vec<LiveLogFeedLinesWrapper>` (`state.rs` `live_log_lines`), evicted
only when the job completed. A long-lived or hung job with chatty output grows
server memory without bound, and each SSE connect (`live_logs_sse`) deep-clones
the whole snapshot, amplifying the spike across concurrent viewers. No
per-message size cap exists beyond axum/tungstenite's transport defaults
(16 MiB frame / 64 MiB message).

Reachability: both ingestion routes sit behind `require_protocol_bearer` on
`protected_apis`, so the vector requires an authenticated runner/job token —
unauthenticated sockets are rejected with 401 before upgrade (covered by
`live_log_websocket_rejects_unauthenticated`). A malicious authenticated runner
can already exhaust the server by other means, so this is a memory-bound
hardening fix (Medium), not a privilege boundary.

## Change

- Replace the per-job `Vec` with a byte-capped `LiveLogBuffer` (`live_logs.rs`):
  - 64 MiB per-job cap, accounting for the wrapper struct, step id, and every
    line string (header + bytes), so heap use is bounded regardless of wrapper
    shape.
  - Oldest wrappers are tail-dropped past the cap; mid-job SSE viewers see the
    retained tail, and the durable step-log blob stays the source of truth.
  - A single wrapper larger than the whole budget is dropped outright: storing
    it would evict the entire tail for one batch, and broadcasting it would
    amplify across every SSE subscriber.
- The cap applies to both ingestion paths (WS + `console_log`), since both route
  through `record_live_log_wrapper`.
- SSE snapshot replay is unchanged — it now clones a bounded buffer.

## Tests

- `buffer_tail_drops_oldest_wrappers_past_cap` — eviction order + byte accounting.
- `buffer_rejects_wrapper_larger_than_whole_budget` — oversized batch dropped,
  tail preserved.
- `buffer_clones_preserve_retained_tail` — snapshot clone / replay contract.
- Existing `live_log_websocket_*` tests updated to the buffer accessors.

## CI

- Local: `cargo fmt --check`, workspace clippy, and full `cargo test --locked
  --workspace` green.
- Control plane (submit-first, run 25): the `ci.yml` `rust` job **passed** on a
  workspace snapshot of this exact tree — fmt-check, clippy `-D warnings`, and
  `cargo test --locked --workspace` all green. `property-tests-fast` stayed
  queued behind a pool stall on the CI host (smolvm golden re-arm failure
  affecting all queued jobs, unrelated to this change; the concurrency
  property profile does not exercise live-log code).

---
## Summary by cubic
Caps per-job live console memory at 64 MiB by replacing the unbounded Vec with a byte-capped `LiveLogBuffer`. Previously we retained all batches until job completion; now we evict the oldest entries to stay within the cap and drop any batch larger than the budget before cloning or broadcasting.

- Applies to both ingestion paths (WS `/ws/live-logs/:job_id` and `TimeLineWebConsoleLog` POST) via `record_live_log_wrapper`.
- Holds the per-job lock across buffer push and broadcast send so live fan-out preserves the retained tail order.
- Oversized batches are rejected pre-clone and not broadcast; a warning is logged.
- State change: `InnerState.live_log_lines` now stores `Mutex<LiveLogBuffer>`; accounting uses allocation capacities and compacts the `VecDeque` after evictions to bound slack; SSE snapshot replay is unchanged but bounded.

<sup>Written for commit 93f5b8e91fa0bb02d8d66deb1c11d833ee692b2e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/preloopdev/preloop/pull/148?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Live logs now retain the newest entries within a 64 MiB limit per job.
  * Oversized log entries are safely rejected, while accepted entries continue streaming in real time.
  * Live-log history remains available when a log view is cloned or resumed.

* **Bug Fixes**
  * Prevented excessive live-log memory usage by automatically evicting older entries when the limit is reached.
  * Ensured rejected entries are not broadcast to live viewers.

### DIFF
--- a/crates/preloop-runner-server/src/lib_tests.rs
+++ b/crates/preloop-runner-server/src/lib_tests.rs
@@ -2276,11 +2276,11 @@ async fn live_log_websocket_accepts_bearer_and_stores_lines() {
                 let inner = state.inner.lock().await;
                 if let Some(job_lines) = inner.live_log_lines.get("job-live") {
                     let wrappers = job_lines.lock().await;
-                    if wrappers.len() == 1 {
-                        assert_eq!(wrappers[0].step_id, "step-1");
-                        assert_eq!(wrappers[0].start_line, 1);
-                        assert_eq!(wrappers[0].count, 2);
-                        assert_eq!(wrappers[0].value, vec!["hello", "world"]);
+                    if wrappers.lines.len() == 1 {
+                        assert_eq!(wrappers.lines[0].step_id, "step-1");
+                        assert_eq!(wrappers.lines[0].start_line, 1);
+                        assert_eq!(wrappers.lines[0].count, 2);
+                        assert_eq!(wrappers.lines[0].value, vec!["hello", "world"]);
                         break;
                     }
                 }
@@ -2361,8 +2361,8 @@ async fn live_log_websocket_survives_malformed_payload() {
             let inner = state.inner.lock().await;
             if let Some(job_lines) = inner.live_log_lines.get("job-malformed") {
                 let wrappers = job_lines.lock().await;
-                if wrappers.len() == 1 {
-                    assert_eq!(wrappers[0].value, vec!["survived"]);
+                if wrappers.lines.len() == 1 {
+                    assert_eq!(wrappers.lines[0].value, vec!["survived"]);
                     break;
                 }
             }
--- a/crates/preloop-runner-server/src/live_logs.rs
+++ b/crates/preloop-runner-server/src/live_logs.rs
@@ -1,5 +1,82 @@
 use super::*;
 
+/// Per-job live console feed buffer.
+///
+/// The ingestion paths (the live-log WebSocket and the `TimeLineWebConsoleLog`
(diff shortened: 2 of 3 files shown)


Bnjoroge1:
@cubic-dev-ai review

PR #201, opened 2026-08-30
feat(cli): make `logs --job/--step` work and add `--follow`

Bnjoroge1:
## What & why

`preloop logs --job X --step N` advertised both filters in `--help`, but they were dead on arrival:

```
no filter      : 33153 bytes
--job bogus    : 33153 bytes
--step 999     : 33153 bytes
>>> IDENTICAL - filters silently ignored
```

`LogsArgs` declared `job`/`step` and `cmd_logs` never read them; `get_run_logs` took no query extractor at all. Any value returned the whole merged run log, exit 0, no warning. The only coverage was a clap parse assertion (`logs_filters`), which validated argument wiring rather than filtering — which is why it went unnoticed.

This makes both filters real and adds `-f`/`--follow`.

**Server** — `GET /api/v1/runs/:run_id/logs` accepts `job` and `step`:

- `job` matches the workflow job key **or** the agent job UUID, reusing the rule `live_log_key_for_job` already applies, so one value works on both log surfaces.
- `step` is a 1-based index into the job's steps in execution order, matching `preloop debug --from`.

The three log tiers are now named (`JobLogs::{Merged,Steps,Live}`) instead of being resolved inline, because only two of them can answer a step query:

- A job whose runner uploaded one merged `job-logs.txt` has **no recoverable step boundaries** → `409`, rather than returning the whole job under a step's name.
- Step numbering restarts per job, so `step` without `job` in a multi-job run → `400` naming the candidates.

**CLI** — `-f`/`--follow` streams a job's live console feed and exits when the run reaches a terminal state. It consumes SSE rather than polling and diffing the whole log: the feed already replays its retained buffer before going live, so a late follower still sees earlier output.

That required a native-bearer route. The existing `/api/v1/runs/:run_id/jobs/:job_id/logs/live` sits in the runner-protocol auth group and `401`s the CLI's token:

```
$ curl -H "Authorization: Bearer <native>" .../jobs/build/logs/live
401 {"error":"runner or job protocol token required"}
```

`/api/v1/runs/:run_id/logs/live` exposes the same stream to the native principal. This grants no new information — that principal can already read the full log via `/logs`. Both handlers share one `live_log_stream` so they cannot drift, and job selection lives server-side (defaults a single-job run, names candidates otherwise) so `?job=` means the same thing streaming or not.

`--follow` with `--step` is refused: the feed carries whole steps as they stream, so honoring both is impossible.

Also corrects the OpenAPI entry for this route, which claimed `text/html` "Rendered log page" for a handler that has always returned `text/plain`.

## Protocol surface

Everything here is on the native `/api/v1/…` management surface. No registration or `/_apis/…` shapes, no broker or NDJSON events, no Twirp results/artifact payloads, no check-run or OAuth behavior.

`live_logs_sse` (runner-protocol group) had its body extracted into `live_log_stream`, but its route, auth, and emitted bytes are unchanged — the refactor is behavior-preserving and covered by the pre-existing live-log tests.

## Required gates

- [x] `just test-ci` passes locally — **with one pre-existing exception, see below**
- [ ] If protocol/wire shapes changed: validated against the official runner — **N/A**, no protocol change
- [ ] Property tests extended — **N/A**, log retrieval has no property suite (those cover concurrency, scheduling, and events)
- [x] New wire fields/events are additive and serde-defaulted — `RunLogsQuery` and `LiveLogsQuery` fields are `Option` + `#[serde(default)]`; an unfiltered request is byte-identical to before
- [x] No secrets, credentials, or internal/deployment-specific paths introduced

### Pre-existing `zizmor` failure, not from this change

`just test-ci` fails at the `zizmor` step on stale `Swatinem/rust-cache` pin comments:

```
warning[ref-version-mismatch]: action's hash pin has mismatched or missing version comment
  --> .github/workflows/ci.yml:49:78
      Swatinem/rust-cache@49a0bdc70d2e1b713ca9e2869b211fcce03d3c1c # v2
                                                                    ^^ points to commit 6323deb102c3
```

This diff touches **zero** workflow files. I stashed all changes and reproduced the identical failure on clean `upstream/main`, so it arrived with the commits that added the check (`661952ba`). It will block CI on `main` until fixed separately.

Every other component passes:

| Gate | Result |
|---|---|
| `just fmt-check` | pass |
| `just clippy` (`-D warnings`) | pass |
| `PROPTEST_CASES=8 cargo test --locked --workspace` | pass (`rc=0`) |
| `just conform` | `conform: PASS` |

## Verification performed

New tests (13):

- `log_run_logs_job_filter_returns_only_that_job` — filter excludes the other job; unfiltered still merges both in request order
- `log_run_logs_job_filter_accepts_agent_job_uuid` — both identifiers work, matching the live feed
- `log_run_logs_unknown_job_is_404_not_whole_run` — asserts the 404 body does not leak the unfiltered log
- `log_run_logs_step_filter_selects_one_step` — steps 1/2/3 return their own content
- `log_run_logs_step_out_of_range_is_404`, `log_run_logs_step_zero_is_rejected_as_one_based`
- `log_run_logs_step_on_merged_upload_is_conflict` — asserts the 409 body does not return the unsplit log
- `log_run_logs_step_without_job_in_multi_job_run_is_ambiguous` — error names both jobs
- `log_run_logs_step_filter_works_on_live_console_blocks` — numeric console-log order, so 2 precedes 10
- `live_run_logs_native_route_{requires_job_when_ambiguous,rejects_unknown_job,defaults_single_job_run,rejects_unauthenticated}`
- CLI: `logs_filters_reach_the_query_string` closes the gap the parse-only test left — it asserts the wire query, not just that clap accepted the flag

End-to-end against a throwaway server on an isolated `PRELOOP_HOME`:

| Case | Result |
|---|---|
| `--job build` / `--job test` | `200`, only that job |
| `--job bogus-job` | `404` ``job `bogus-job` not found in this run`` |
| `--step 1` (2 jobs) | `400` ``` `step` needs `job` when a run has 2 jobs: build, test ``` |
| `--job build --step 99` | `404` ``job `build` has 0 steps so far`` |
| `--job build --step 0` | `400` `step is 1-based` |
| `-f --job build`, live run | held the stream 10s (`rc=124`) |
| `-f` on a terminal run | exited `rc=0` in 3.1s via terminal-state detection |
| `-f --step 1` | refused before connecting |

## Checklist

- [x] Docs updated — `docs/cli_reference.md` documents all three flags, the per-job step numbering rule, and the merged-log limitation

### User-facing summary

`preloop logs --job` and `--step` now filter instead of silently returning everything, and `preloop logs -f` tails a running job.

---
## Summary by cubic
`preloop logs --job` and `--step` previously returned the entire run log; they now filter server-side, and `-f`/`--follow` tails a selected job's live output until that job finishes. Follow mode replays retained output, ends when the job reaches a terminal state, and falls back to the durable log when no in-memory snapshot exists.

- Job filters accept the workflow job key or agent job UUID; unknown and cross-run jobs are rejected.
- Step numbers follow workflow order via runner step IDs, preserve duplicate step names, and require `--job` for multi-job runs.
- Step queries on merged logs return `409`; `--follow --step` is refused because live feeds are job-scoped.
- Adds an authenticated native SSE route that closes feeds on completion and reopens them for retries; a completed job with no snapshot falls back to the durable log.
- Updates CLI documentation, OpenAPI responses (including the `text/html` to `text/plain` correction), SSE parser coverage, and CI/Node dependency pins.

<sup>Written for commit afbee26bcde65743ad7246ea26f5ad1e00e891e8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/preloopdev/preloop/pull/201?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added job and step filters to run logs.
  - Added `--follow`/`-f` to stream live job logs.
  - Added an authenticated live-log API endpoint.
  - Live streams now close when jobs finish and support retries and late connections.
- **Bug Fixes**
  - Improved log retrieval across merged, per-step, and live sources.
  - Preserved separate logs for steps with duplicate names.
  - Added clearer errors for invalid, ambiguous, or unavailable filters.
- **Documentation**
  - Expanded CLI log documentation with filtering, live-follow behavior, limitations, and examples.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,7 +46,7 @@ jobs:
           toolchain: "1.97"
           components: rustfmt,clippy
 
-      - uses: Swatinem/rust-cache@49a0bdc70d2e1b713ca9e2869b211fcce03d3c1c # v2
+      - uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2
 
       - name: Install lld linker
         run: apt-get update && apt-get install -y --no-install-recommends lld
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -25,7 +25,7 @@ jobs:
           persist-credentials: false
           lfs: true
       - uses: dtolnay/rust-toolchain@4360b52568e2003a75bf9bc1d59f33a8e3fc893c # stable
-      - uses: Swatinem/rust-cache@49a0bdc70d2e1b713ca9e2869b211fcce03d3c1c # v2
+      - uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2
       - name: Replay official runner flows
         run: bash ./benchmarks/conformance/run.sh
       - name: Upload light report
--- a/crates/preloop-cli/src/main.rs
+++ b/crates/preloop-cli/src/main.rs
@@ -4382,15 +4382,10 @@ mod tests {
     fn live_log_frame_prints_only_console_lines() {
         let frame = "event: live-log\ndata: {\"stepId\":\"s1\",\"startLine\":1,\"count\":2,\
                      \"value\":[\"first\",\"second\"]}";
-        // Captured behavior: the payload's `value` lines are the output, and the
-        // `event:` name is not.
-        let wrapper: preloop_gha_protocol::LiveLogFeedLinesWrapper = frame
-            .lines()
-            .find_map(|line| line.strip_prefix("data:"))
-            .map(|payload| serde_json::from_str(payload.trim_start()).unwrap())
-            .expect("frame carries a data payload");
-        assert_eq!(wrapper.value, vec!["first", "second"]);
-        assert_eq!(wrapper.step_id, "s1");
+        // The production parser reports whether this frame emitted console
+        // lines; the `event:` line itself carries no output.
+        assert!(print_live_log_frame_bytes(frame.as_bytes()).unwrap());
+        assert!(!print_live_log_frame_bytes(b"event: live-log").unwrap());
     }
 
     #[test]
--- a/crates/preloop-cli/src/main.rs
+++ b/crates/preloop-cli/src/main.rs
@@ -3532,7 +3532,7 @@ async fn cmd_logs(args: LogsArgs) -> anyhow::Result<()> {
 
 /// Fetch a run's overall execution status.
 ///
-/// `None` when the field is absent or unrecognized, which the caller treats as
+/// `None` when the field is absent or unrecognized, which callers treat as
 /// "keep following" rather than a reason to stop.
 async fn run_status(
     client: &reqwest::Client,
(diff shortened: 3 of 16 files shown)


Bnjoroge1:
@codeRabbit review

PR #204, opened 2026-08-30
fix(node-externals): manifest validation + checksum verification + refresh hooks

Bnjoroge1:
## Summary

Fixes the stale-cache root cause from the audit: externals installation skipped on binary existence alone, so pin bumps never refreshed existing hosts/bundles (production stayed on Node 20.19.0/24.3.0), and downloads had no integrity verification (`curl | tar`).

### Changes
- **Single source of truth**: both `preloop-orchestrator` and `preloop-runner` read node pins from `versions.toml` at build time (`build.rs`), with fallback to the previous pins if keys are absent. No more hardcoded copies in `configure.rs`.
- **`preloop-node.json` manifest** per installed runtime: `{runtime, version, platform, archive_sha256, source}`. Cache validation = manifest present AND version matches expected AND `bin/node --version` agrees (`is_valid_externals_dir`). Stale/missing/mismatched → re-materialize that runtime atomically.
- **Checksum verification**: pinned SHA-256 table compiled from `versions.toml` (`[node_externals.sha256]`) is authoritative; `SHASUMS256.txt` fetched from nodejs.org is verified as belt-and-braces. Orchestrator's `curl | tar` pipe replaced with download→verify→extract; runner extracts from verified in-memory bytes.
- **Refresh hooks**: `preloop update` (both self-replace and Linux runner-bundle install) invalidates managed externals whose manifest version ≠ newly compiled-in pin, so the next engine start re-materializes. `ensure_host_externals` and bundle materialization now validate manifests instead of checking existence.

### Verification
- Unit tests (14): stale manifest triggers re-download; digest-mismatched tarball rejected with no install; fresh valid cache not re-downloaded; manifest write/read round-trip.
- `cargo fmt --check`, `cargo clippy -D warnings`, `cargo test -p preloop-orchestrator -p preloop-runner`: green. E2E submitted (3-job run) before push.

📌 Depends on #203 for the actual pin values; correct without it (falls back to old pins).

---
## Summary by cubic
Stops Node externals from silently reusing stale or corrupt runtimes. Every install now requires manifest validation and SHA256 checksum verification; anything stale, mismatched, or unverified is re-downloaded instead of reused.

**Validation and checksums**
- Each runtime dir requires a `preloop-node.json` manifest and a matching `bin/node --version`; anything else triggers a re-download.
- Tarballs verify against a pinned SHA from `versions.toml` and `SHASUMS256.txt`; downloads fail closed when no trusted checksum exists.
- Non-native runtimes (such as `linux-arm64` on a macOS host) validate via manifest metadata since the foreign binary can't be executed.

**Refresh**
- `preloop update` deletes stale manifests in the host, runner bundle, and `PRELOOP_RUNNER_EXTERNALS` paths.
- Node versions and SHA pins are compiled from `versions.toml`, and `hex` was added as a workspace dependency.

<sup>Written for commit 3a085f1166484834c33b0eb7ffd94fb84d43c00a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/preloopdev/preloop/pull/204?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added reliable Node.js runtime management across supported platforms.
  * Added provenance tracking for installed runtimes.
* **Bug Fixes**
  * Automatically replaces stale, invalid, malformed, or mismatched cached runtimes.
  * Prevents unverified, incomplete, or tampered downloads from being installed.
* **Improvements**
  * Reuses valid cached runtimes while refreshing outdated installations.
  * Applies consistent pinned runtime versions and checksums across workflows.
  * Improved validation of native and cross-platform runtime bundles.
  * Supports checksum verification using trusted archive metadata when available.


Bnjoroge1:
@coderabbit review

PR #206, opened 2026-08-30
feat(cli): derive local event context; explain trigger mismatches

Bnjoroge1:
## What & why

`preloop run` could not run a path-filtered workflow at all, against the **drop-in workflows** guarantee in AGENTS.md. Three separate problems.

### 1. The CLI never supplied a changed-file list

`on.<event>.paths` / `paths-ignore` needs one, and the server correctly fails closed without it. But the CLI set neither `changed_paths` nor `changed_paths_known` — anywhere — and `--base` only feeds `base_ref` (branch filtering). So every path-filtered workflow died on submit unless the user hand-wrote a webhook payload:

```
$ preloop run -f .github/workflows/runner-conformance.yml --event=pull_request --job runner-light
Error: 400 {"error":"workflow path filters require a complete changed-file list"}
```

The CLI now derives it from git, beside the existing `default_local_activity_type`: the merge-base diff against a base, **union uncommitted changes** — the server snapshots the working tree, so a filter judged only on committed history would ignore the edit under test.

Base resolution prefers `--base`, then the branch's tracking ref, then each remote's default, then local `main`/`master`. Each candidate must **share history with HEAD** — this repo's `origin` is a fork whose `origin/HEAD` resolves fine but has no merge base, which produced `fatal: no merge base` and an empty diff during testing. When nothing usable is found the change set stays *unknown* rather than empty, because an empty **known** list makes every `paths:` filter reject the run.

Also derives `filter_branch` for PR events. GitHub applies `on.pull_request.branches` to the PR's **target** branch; without it a local PR run filtered on the checked-out branch, so a workflow gated to `branches: [main]` never matched from a feature branch.

**Deliberately not derived:** PR number, actor, labels, review state, and `workflow_run` upstream results have no local truth. Guessing them would flip `if:` conditions on fabricated data, so they stay absent and `--payload` remains the way to supply them.

### 2. "workflow does not match event X" said nothing actionable

That message could mean the event is undeclared, or declared but excluded by any of five filter axes, with no way to tell which. `matches_with_context` now delegates to `match_event`, which returns a `TriggerMismatch` naming the axis and the offending values; the boolean is `.is_ok()` on the same call, so the two cannot drift. The server appends the flag that changes each outcome:

```
workflow does not run for event `pull_request`: the workflow declares only:
push, workflow_dispatch. Pass `--event push` or `--event workflow_dispatch`.

workflow does not run for event `pull_request`: activity type `synchronize` is
not accepted; the workflow accepts: opened, reopened. Supply one with
`--payload` containing {"action": "opened"}, or pick an activity type the
workflow accepts.

workflow does not run for event `push`: branch `Bnjoroge/local-path-filters` is
excluded by branches: [release-*]. Check out a matching branch or tag, or pass
`--base <REF>` for pull_request events (the branch filter applies to the PR's
target branch).

workflow does not run for event `pull_request`: all 1 changed file(s) match
paths-ignore: [**.md]. Change a file the filter selects, or pass `--base <REF>`
to diff against a different base.
```

The path-filter error is likewise rewritten to name what to pass, for both CLI and API callers.

### 3. `paths-ignore: ["**/*.md"]` never ignored root-level docs

`**` matches zero or more of any character, so the literal `/` after it still needs a real separator: `**/*.md` skips `docs/x.md` but **not** `README.md`. GitHub documents `**.md` for "any file in the repository with the .md extension".

With `AGENTS.md`, `README.md`, and `CONTRIBUTING.md` all at the repo root, a docs-only PR was running full CI — including the 10-minute self-hosted `runner-conformance` job. Fixed in all four workflows (`ci`, `conformance`, `runner-conformance`, `zizmor`).

These semantics had **zero test coverage**, so preloop agreed with GitHub's documented rule by construction rather than verification. Both forms are now pinned.

## Protocol surface

No registration or `/_apis/…` shapes, no broker or NDJSON events, no Twirp payloads, no check-run or OAuth behavior. `changed_paths` / `changed_paths_known` / `filter_branch` are existing `WorkflowSubmission` fields that the webhook dispatcher already populates (`github.rs`); this teaches the local CLI to populate them the same way. Trigger matching keeps identical accept/reject behavior — `match_event` is the same logic returning a reason instead of a bool, with a test pinning the two to agree.

## Required gates

- [x] `just test-ci` passes locally — **with one pre-existing exception, see below**
- [ ] Protocol/wire shapes validated against the official runner — **N/A**, no protocol change
- [ ] Property tests extended — **N/A**, trigger matching has no property suite (those cover concurrency, scheduling, events)
- [x] New wire fields are additive and serde-defaulted — no new fields; all three are existing `#[serde(default)]` members
- [x] No secrets, credentials, or internal/deployment-specific paths introduced

### Pre-existing `zizmor` failure, unchanged by this PR

`just test-ci` fails at `zizmor` on stale `Swatinem/rust-cache` pin comments in `ci.yml` and `conformance.yml`:

```
warning[ref-version-mismatch]: action's hash pin has mismatched or missing version comment
      Swatinem/rust-cache@49a0bdc70d2e1b713ca9e2869b211fcce03d3c1c # v2
                                                                    ^^ tag points to commit 6323deb102c3
```

Identical count before and after my edits (92 findings, 2 medium), verified against clean `upstream/main`. I looked it up: that SHA matches **no tag** in `Swatinem/rust-cache`, so correcting it means *bumping the action* — a functional CI change I'm not folding into a path-filter PR. Happy to send it separately.

Other gates:

| Gate | Result |
|---|---|
| `just fmt-check` | pass |
| `just clippy` (`-D warnings`) | pass |
| `PROPTEST_CASES=8 cargo test --locked --workspace` | pass (`rc=0`) |
| `just conform` | `conform: PASS` |

## Verification performed

End-to-end, isolated `PRELOOP_HOME`, before/after with the old binary against the same server:

| Case | Old CLI | This PR |
|---|---|---|
| `--event=pull_request --job runner-light` (path-filtered, no payload) | `400` needs changed-file list | run created |
| PR gated `branches: [main]`, from a feature branch | `400` branch excluded | run created |
| `paths-ignore: ["**.md"]`, only `README.md` changed | — | correctly **skipped**, reason given |
| `paths: ["src/**"]`, only docs changed | — | correctly **skipped**, reason given |
| undeclared event | generic mismatch | lists `--event push` / `--event workflow_dispatch` |
| activity type rejected | generic mismatch | names accepted types + `--payload` snippet |

Filters are still *evaluated*, not bypassed — the skip cases above are the proof.

New tests (17):

- `glob_match_double_star_requires_a_real_separator_before_a_literal_slash`, `glob_match_single_star_does_not_cross_separators` — pin `**` vs `**/` vs `*`
- `match_event_agrees_with_matches_with_context` — the no-drift guard between predicate and reason
- `match_event_reports_undeclared_event_with_the_declared_list`, `..._missing_and_rejected_activity_types`, `..._default_pr_activity_types`, `..._ref_and_path_filters`, `..._all_paths_ignored` (including that a *mixed* change set still runs), `..._unmatched_upstream_workflow`
- CLI: `porcelain_path_handles_status_columns_renames_and_quotes` (renames keep only the destination; `core.quotePath` quotes stripped), `only_file_list_events_derive_changed_paths`, `explicit_payload_paths_suppress_derivation`, `filter_branch_derives_pr_target_not_head`

One design fix came from a test failure rather than review: `filter_branch` initially required the base to resolve locally, which breaks shallow clones and un-fetched bases. An explicit `--base` now names the target branch regardless.

## Checklist

- [x] Docs updated — `docs/cli_reference.md` gains a "What a local run derives" table, the base-resolution order, and an explicit list of what is *not* synthesized and why

### User-facing summary

`preloop run` now runs path-filtered and `pull_request` workflows without a hand-written payload, and a workflow that does not trigger explains which filter rejected it and which flag to pass. `paths-ignore: ["**/*.md"]` → `["**.md"]` so docs-only PRs stop running full CI.

---
## Summary by cubic
`preloop run` now derives path and branch filter context from git, so path-filtered workflows run locally without a hand-written `--payload`, and trigger rejection errors name which filter failed and what to pass.

**New Features**
- Derives the changed-file list from the merge-base diff against a resolved base plus uncommitted changes, using NUL-delimited git output for accurate unicode and escaped path handling.
- Resolves the diff base preferring `--base`, then the branch's tracking ref, each remote's default branch, then local `main`/`master`, skipping candidates that do not share history with `HEAD`.
- Leaves the change set unknown when nothing resolves, and suppresses derivation when `--payload` already carries `paths` or `commits`.
- Derives `filter_branch` for `pull_request`/`pull_request_target` from the PR's target branch, preserving internal branch slashes; an explicit `--base` names it even when the ref is not fetched locally.
- `match_event` returns a `TriggerMismatch` naming the rejected axis and values; `matches_with_context` delegates to it so the two cannot drift.
- The server appends the flag that changes the outcome and reads `workflow_run.name` from the payload to satisfy upstream workflow triggers.
- Pins the default PR activity types to `opened`, `synchronize`, `reopened`.
- PR number, actor, labels, review state, and upstream run names are deliberately not derived; workflows needing them still require `--payload`.

**Bug Fixes**
- Switched `paths-ignore` from `**/*.md` to `**.md` in `ci`, `conformance`, `runner-conformance`, and `zizmor` so root-level markdown files are ignored and docs-only PRs stop running full CI.
- Pinned `**` vs `**/` vs `*` glob semantics with tests.
- Improved tag vs branch mismatch diagnostics so a branch push against tag-only filters (and vice versa) is explained correctly.

<sup>Written for commit 7176f88c23aa8edfcbaa2a7b526e32815d50f105. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/preloopdev/preloop/pull/206?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Local `preloop run` now derives changed files, target branches, and pull request context from the current Git checkout, enabling path and branch filters to work during local runs.
  - Workflow trigger matching now reports specific mismatch reasons and actionable guidance.
  - Upstream workflow names are recognized from workflow-run data when needed.

- **Bug Fixes**
  - Improved handling of Git paths and change sets, including uncommitted changes and quoted filenames.
  - Corrected pull request activity matching for synchronization events.

- **Documentation**
  - Added details about context and change detection for local runs.


Bnjoroge1:
@cubic-dev review

PR #210, opened 2026-09-01
fix(debug): report workflow step positions without lifecycle steps

Bnjoroge1:
## What & why

Debug output counted runner-generated steps as workflow steps, so `preloop debug` reported positions a user could not find in their YAML, and `--from N` selected the wrong step. `Set up job`, `Pre`/`Post` action hooks, container lifecycle steps, host hooks and `Complete job` all occupied positions.

This separates the position a user sees from the index the worker retries:

- the resolved runner index is preserved for retry fidelity
- workflow-facing positions count only steps the workflow declared
- `--from` and `--from-start` operate on workflow numbering
- session listings and banners render `workflow step 3/4`

### How a step is classified

A resolved step is synthetic **iff its id is absent from the job request message's step list**. The worker already has that list — `main_steps` in `job_runner.rs` is 1:1 with `job_message["steps"]` — so it is captured on `JobContext` before lifecycle expansion adds anything.

This replaces an earlier prefix/marker heuristic. Prefixes cannot do this job: `job_builder.rs:363-364` passes a user's `id:` through verbatim as the context name, so a workflow declaring `id: __pre_build` had its own step hidden from counts and `--from` selection, with its failure mislabelled as a lifecycle failure. Lifecycle steps parsed from a live payload can also carry a GUID id with markers spread across `raw.__pre`, `raw.__post`, `raw.isPre` and `raw.isPost`, so any prefix list is one shape away from being wrong. Membership has no shapes to miss.

Verified against the official runner in `.runner-watch/golden/v2.336.0/06-multi-step`: the three declared steps echo the message ids back as `external_id`, while `Set up job` and `Complete job` carry ids absent from it.

### Retry-selection fixes

- `--from-start` selected the first workflow step even when it sat *after* the failure. With a leading synthetic failure the worker rejected that index and silently retried the synthetic step — running something other than the command it accepted. It now selects the first workflow step at or before the failure, and explains itself when none has started.
- An unmatched `--from` listed steps after the failure as available, which the retry path rejects. That list now comes from the same bounded set.

## Protocol surface

No `/_apis/…`, broker, Twirp results-service, check-run or OAuth shape is touched. The official runner is unaffected: classification is computed inside the worker from a message it already received, and nothing new is sent to it.

The preloop-native **debug-session** wire shape does change: `StepSummary.is_internal` is renamed to `synthetic`, aligning with the server's `StepKind::Synthetic` for the identical concept (see #212). The field stays `#[serde(default, skip_serializing_if = ...)]`, so a payload without it deserializes as `synthetic: false`. The honest consequence: a session captured before this change would render its lifecycle steps as workflow steps, because the CLI's prefix fallback is deliberately removed. That is acceptable pre-release — sessions are transient control-plane state, not persisted across versions — and it is why the fallback went rather than being scoped to legacy summaries.

## Required gates

- [ ] `just test-ci` passes locally — **partially; see below**
- [x] Protocol/wire shapes: validated against the official runner via the golden capture that establishes which ids the runner echoes back
- [ ] Property tests — not applicable; the touched modules have no property suites
- [x] New wire fields are additive and serde-defaulted
- [x] No secrets, credentials, or deployment-specific paths introduced

`just dogfood` cannot run in this environment: `~/.cache/actions-runner/current` is absent. The `zizmor` `ref-version-mismatch` findings noted when this PR was opened were fixed on `main` in #201 and no longer reproduce.

## Verification performed

```text
cargo fmt --all -- --check                                       clean
cargo clippy --locked --workspace --all-targets -- -D warnings    clean
cargo test -p preloop-runner --lib                               557 passed
cargo test -p preloop-cli --bin preloop                          164 passed
cargo test -p preloop-gha-protocol --lib -- debug_session          6 passed
```

New: `synthetic_classification_follows_declared_step_membership` pins the classification across every shape a lifecycle step can take:

| case | expected |
|---|---|
| declared script step | workflow |
| post hook from an action manifest (`raw.__post`) | synthetic |
| declared step whose author wrote `id: __post_build` | workflow |
| GUID-identified lifecycle step marked only in `raw.isPre` | synthetic |
| host `__hook_job_started` | synthetic |

Rows 3 and 4 are the cases that fail under any prefix list, so the test pins the reason for the design rather than only its outcome.

Also covered: `workflow_positions_hide_runner_lifecycle_steps`, `retry_from_numbers_use_workflow_positions`, `retry_from_accepts_start_number_and_unique_name`, `retry_from_rejects_future_ambiguous_and_conflicting_selections`.

## Checklist

- [ ] Docs — no user-facing doc change needed; `docs/cli_reference.md` already documents `--from` in workflow-step terms, which this makes true. #212 updates the `--step` half of that contract.

## Related

#212 makes the same declared-vs-generated distinction on the server for `preloop logs --step N`, using the same rule against the same source. No file overlap. Together they make `docs/cli_reference.md`'s claim that `--step` "mirrors `preloop debug --from`" true by construction rather than by coincidence.

## Summary by CodeRabbit

* **New Features**
  * Debug sessions now distinguish workflow steps from internal lifecycle, hook, and cleanup steps.
  * Step listings, banners, and retry prompts use workflow-relative numbering for clearer navigation.
  * Retry commands provide improved feedback when selecting unavailable workflow steps or restarting before any workflow step has run.
  * Older sessions remain compatible through legacy step-resolution behavior.


Bnjoroge1:
Rebased onto `main` and addressed all five findings. Pushed as `42f4b90f`.

The classifier findings shared a root cause, so rather than patch each shape I replaced the heuristic with the rule that's actually authoritative.

## Classification is now membership, not prefixes

A resolved step is synthetic **iff its id is absent from the job request message's step list**. The worker already has that list — `main_steps` in `job_runner.rs` is 1:1 with `job_message["steps"]` — so it's captured on `JobContext` before `build_step_list_with_lifecycle` and the hook injections add anything:

```rust
job_ctx.declared_step_ids = main_steps.iter().map(|step| step.id.clone()).collect();
```

```rust
fn is_synthetic_step(step: &Step, declared_step_ids: &HashSet<String>) -> bool {
    !declared_step_ids.contains(&step.id)
}
```

Verified against the official runner in `.runner-watch/golden/v2.336.0/06-multi-step`: the three declared steps echo the message ids back as `external_id`, while `Set up job` and `Complete job` carry ids absent from it.

### @chatgpt-codex-connector (`debug_session.rs:188`), @cubic-dev-ai (`debug_session.rs:186`) — user steps with lifecycle-like names

Valid, and this was the sharpest finding. `job_builder.rs:363-364` passes a user's `id:` through verbatim as the context name, so a workflow declaring `id: __pre_build` had its own step dropped from workflow counts and `--from` selection, with its failure mislabelled as a lifecycle failure.

Rather than track field presence to scope the fallback to legacy summaries, the fallback is **gone**. The CLI now reads the wire flag and nothing else. There are no older sessions to stay compatible with — this is pre-release — so the compatibility branch was carrying only risk. A regression case covers exactly your example.

### @cubic-dev-ai (`steps_runner.rs:142`), @coderabbitai (`steps_runner.rs:141`) — missed lifecycle shapes

Valid. Your suggested patch would have worked, but membership makes the enumeration unnecessary: `raw.__pre`, `raw.__post`, `raw.isPre`, `raw.isPost`, context-name prefixes and GUID-identified lifecycle steps all resolve correctly because none of them appear in the declared list. That removes the class of bug rather than the four instances of it, and there's no fifth shape to discover later.

### @cubic-dev-ai (`steps_runner.rs:130`) — no end-to-end regression test

Valid. Added `synthetic_classification_follows_declared_step_membership`, which asserts the classification across every shape in one table:

| case | expected |
|---|---|
| declared script step | workflow |
| post hook from an action manifest (`raw.__post`) | synthetic |
| declared step whose author wrote `id: __post_build` | workflow |
| GUID-identified lifecycle step marked only in `raw.isPre` | synthetic |
| host `__hook_job_started` | synthetic |

Rows 3 and 4 are the ones that fail under any prefix list, so the test pins the reason for the design and not just the outcome.

## Retry-selection bounds

### @chatgpt-codex-connector (`:488`), @coderabbitai (`:489`), @cubic-dev-ai (`:488`) — `--from-start` before the first workflow step

Valid. `--from-start` selected the first workflow step even when it sat *after* the failure, so with a leading synthetic failure the worker rejected the target and silently retried the synthetic step — executing something other than the command it accepted. It now selects from `workflow_steps_before_failure` and fails with an explanation when no workflow step has started:

```text
no workflow step has started yet: runner lifecycle step 1/6 (not a workflow step) failed before the first one
```

### @cubic-dev-ai (`:540`) — unmatched `--from` advertising unreachable steps

Valid, applied as suggested. The available-steps list now comes from `workflow_steps_before_failure`, so the guidance matches what the retry path accepts.

## Vocabulary

Renamed the wire field `is_internal` → `synthetic`, matching the server-side `StepKind::Synthetic` for the identical concept (see #212). Two names for one idea on surfaces a user reads side by side would drift; `synthetic` also reads better than `internal`, which suggests "private implementation detail" rather than "the runner generated this".

## Relationship to #212

#212 does the same declared-vs-generated split on the server, for `preloop logs --step N`. No file overlap with this PR. Both now classify by the same rule against the same source, so `docs/cli_reference.md`'s claim that `--step` "mirrors `preloop debug --from`" holds by construction instead of by coincidence — two independent heuristics would have let it drift again.

## Verification

```text
cargo fmt --all -- --check                                       clean
cargo clippy --locked --workspace --all-targets -- -D warnings    clean
cargo test -p preloop-runner --lib                               557 passed
cargo test -p preloop-cli --bin preloop                          164 passed
cargo test -p preloop-gha-protocol --lib -- debug_session          6 passed
```

Still outstanding from the original description, unchanged by this push: `just dogfood` cannot run here because `~/.cache/actions-runner/current` is absent. The `zizmor` `ref-version-mismatch` findings noted originally were fixed on `main` in #201 and no longer reproduce.

PR #212, opened 2026-09-01
fix(server): own step identity and order from the job request message

Bnjoroge1:
## What & why

`preloop logs --job X --step N` could return the wrong step's log, and repeated
step names could lose a step from the run record.

Both came from the same root cause: the server had no record of a job's declared
step order, so it reconstructed structure from whatever the runner happened to
report and from whatever landed on disk. Specifically it sorted `step-*.txt` by
filesystem modification time — which records when a blob was uploaded, not when
a step ran — and it matched steps by display name, which is not unique.

This makes the job request message the source of truth. Each attempt's step
manifest is built from `AgentJobRequestMessage.steps` and recorded at dispatch,
before the runner reports anything. Runner reports only reconcile into it, keyed
by `external_id`.

Verified against the official runner before writing any code
(`.runner-watch/golden/v2.336.0/06-multi-step`):

```text
job message steps (in order)      runner-reported
[0] d9cb9a38…                     number=2  d9cb9a38…  'Run echo first'      in_message=true
[1] 6baf5458…                     number=3  6baf5458…  'Run echo "VAL=$VAL"' in_message=true
[2] 62f67d20…                     number=4  62f67d20…  'Run echo line1'      in_message=true
                                  number=1  23c0d0c1…  'Set up job'          in_message=false
                                  number=5  456c6a95…  'Complete job'        in_message=false
```

Three things this establishes:

1. `external_id` **is** `TaskStep.id` for declared steps — the identity chain holds end to end.
2. Steps the runner generates carry ids absent from the message, so membership in
   the manifest classifies a record as `workflow` or `synthetic`. Not id shape,
   not display name.
3. The runner's `number` is offset by synthetic steps (declared step 1 reports as
   `number: 2`), so it cannot drive `--step`.

GitHub's own blob path is `logs/steps/step-logs-<external_id>.txt`, so keying
durable logs by that id matches upstream.

### User-facing behavior

- `--step N` selects the Nth **declared** step. `Set up job`, `Pre`/`Post` hooks,
  container lifecycle and `Complete job` keep their logs in the whole-job output
  but no longer shift the numbering read off the YAML.
- Two steps sharing a `name:` stay distinct.
- When the declared order genuinely cannot be recovered, the request fails with
  `409` instead of returning a neighbouring step under the requested step's name.

### Design notes

Manifests are keyed by `agent_job_id`, not by job key. `build_task_step`
(`job_builder.rs:844`) mints a fresh `Uuid::new_v4()` per dispatch, so a
job-scoped map would overwrite the mapping an earlier attempt's `step-<id>.txt`
blobs are still named after, making that attempt's logs unreachable.

Step records get a `job_steps` table rather than riding in `runs.record_blob`.
That blob carries the workflow YAML, the event payload and the github context,
and `store_run_event` reseals all of it on every run event. `MetaSnapshot` is no
better — `store.rs:511-518` documents that every field in it is cloned and sealed
on each `store_meta_only` call, and that path fires per step-log upload. Rows are
upserted by `(agent_job_id, step_id)` with no per-run delete, so a step
transition writes one small row. `kind` and `workflow_index` stay plaintext so
`--step` resolves through an indexed query; only the display name is sealed,
since it comes from user YAML.

`JobDetail` gains an explicit `job_id`. Previously `name` served as both the job
key and the display name — the run projection overwrote it with the display name
and then looked up by either, which is the same identify-by-name mistake one
level up.

`context_name` is persisted because it is the one identity stable across runs of
the same workflow; the step id is per-dispatch.

## Protocol surface

Runner-facing wire shapes are unchanged. `WorkflowStepsUpdate` and the timeline
PATCH handlers were rewritten internally, but their request/response payloads
and status codes are byte-identical — the conformance replay reports
`WorkflowStepsUpdate` as `1 call, 200 vs 200` with identical bodies and no
missing endpoints on either side.

`GET /api/v1/runs/{run_id}` (native API, not the runner protocol) gains additive
serde-defaulted fields on step records: `kind`, `workflow_index`,
`runner_number`, `context_name`, plus `job_id` on job details.

## Required gates

- [ ] `just test-ci` passes locally — **NO, see caveat below**
- [x] Validated against the official runner: golden replay
      (`v2.336.0/06-multi-step`) established the identity contract before
      implementation, and the conformance replay shows the rewritten handlers
      unchanged on the wire
- [ ] Property tests — not applicable; the touched subsystems
      (`results_twirp`, `timeline_logs`, log resolution) have no property suites
- [x] New wire fields are additive and serde-defaulted
- [x] No secrets, credentials, or deployment-specific paths introduced

### Caveat on `just test-ci`

`conform` reports 6 failing scenarios. I did not establish a pre-change baseline,
so I am **not** claiming these predate this branch. What the reports do show:
endpoint counts, status codes and request/response bodies all match official,
no endpoints missing on either side, and the diffs present are header-key
differences (`activityid`, `x-tfs-processid` and other Azure infrastructure
headers preloop does not emit). A baseline `just conform` on `main` settles it in
one run and should happen before merge.

`just dogfood` could not run: `~/.cache/actions-runner/current` is absent in this
environment.

## Verification performed

```text
cargo test -p preloop-runner-server --lib      649 passed, 0 failed, 3 ignored
cargo fmt --all -- --check                     clean
cargo clippy --locked --workspace --all-targets -- -D warnings   clean
```

New coverage for cases that previously had none:

- `step_manifests_are_scoped_per_job_attempt` — a re-dispatch mints a fresh
  manifest and both attempts' logs stay resolvable
- `step_manifests_survive_a_restart` — manifests round-trip the `job_steps`
  table, and `--step 2` still resolves against a fresh `AppState` over the same
  state dir
- `log_run_logs_step_filter_uses_workflow_step_ids` — now also asserts a
  synthetic upload does **not** become step 2

Six existing tests were rewritten. All six failed for one reason: they
fabricated `external_id`s that were never in the job message, so reconciliation
correctly classified them as synthetic. The duplicate-name test demonstrated the
fix while failing — two declared steps stayed distinct with separate ids; it saw
4 records only because it invented 2 more. They now report under real manifest
ids, which is what the runner does.

## Checklist

- [x] Docs updated — `docs/cli_reference.md` describes where `--step` numbering
      comes from, which steps are excluded, and when it returns `409`

## Related

Overlaps conceptually with #210, which introduces the same
declared-vs-generated distinction on the debug/runner side. No file overlap.
Landing this first settles the shared vocabulary (`synthetic`) and lets #210
classify by manifest membership instead of prefix heuristics.

---
## Summary by cubic
Makes the job request message the source of truth for step identity and order, so `preloop logs --job X --step N` no longer returns a neighboring step's log or drops repeated-name steps. Each dispatch owns an attempt-scoped manifest seeded before the runner reports and reconciled by `external_id`; runner wire payloads are unchanged, and native run responses add serde-defaulted step metadata and an explicit `job_id`.

**Bug Fixes**

- `--step N` counts declared workflow steps; synthetic runner steps stay in whole-job output without shifting positions.
- Completions, live-log follow, and compatibility requests resolve the reporting or newest attempt, not the oldest.
- Timeline reports keep untyped and `Step` records, exclude container and job records, and order all records by execution position.
- Requests return `409` when order is unavailable, per-step blobs have not landed, or only live console blocks exist.
- Whole-job logs, run projections, and run-list responses use the same ordered step records, including AzDO synthetic steps.
- The conformance replay's 6 failing scenarios reproduce on `main` with the same count.

**Persistence**

- Adds per-attempt `job_steps` storage in SQLite and Postgres, rebuilds missing manifests from the persisted request at startup, and drops manifests for retired requests.
- Revision checks prevent stale writes; restored revisions seed the next write, and the schema guard distinguishes stale intermediate databases from real failures like a locked database.
- Both backends create `job_steps` in one migration; dev databases stamped 4 by an earlier build of this branch fail startup with an actionable error.

<sup>Written for commit 97ede6cac662d6ca40d0f4a66684cf853c6c35da. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/preloopdev/preloop/pull/212?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved step status and completion reporting for retries, dynamic jobs, and runner-added steps.
  - Ensured logs and step lists follow execution order, including timeline-reported steps.
  - Prevented synthetic runner steps and parent job records from occupying workflow step positions.
  - Improved log filtering across restarts and repeated job attempts.
  - Added validation for malformed or unavailable step reports.

- **Documentation**
  - Clarified `preloop logs --step` numbering, job requirements, runner-generated steps, duplicate names, and failure conditions.

- **Persistence**
  - Retained step manifests and statuses across restarts for each job attempt.
  - Prevented delayed updates from overwriting newer step results.

### DIFF
--- a/crates/preloop-gha-protocol/src/lib.rs
+++ b/crates/preloop-gha-protocol/src/lib.rs
@@ -696,6 +696,15 @@ pub struct JobCompletion {
     pub run_id: RunId,
     /// Job id.
     pub job_id: JobId,
+    /// The attempt this completion belongs to, when the caller resolved one.
+    ///
+    /// A job can be dispatched more than once, and each dispatch mints its own
+    /// step ids. `(run_id, job_id)` names the logical job, not the attempt, so
+    /// it cannot decide which attempt's step records a report belongs to.
+    /// Callers that already hold the request record supply this; the server
+    /// falls back to the newest attempt when it is absent.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub agent_job_id: Option<uuid::Uuid>,
     /// Final status.
     pub status: ExecutionStatus,
     /// Outputs captured by the runner.
--- a/crates/preloop-runner-server/src/bootstrap.rs
+++ b/crates/preloop-runner-server/src/bootstrap.rs
@@ -397,6 +397,11 @@ pub(crate) async fn reap_once(shared: &Arc<SharedState>) {
                     JobCompletion {
                         run_id,
                         job_id: job_id.clone(),
+                        // The lease that expired names the attempt exactly.
+                        agent_job_id: inner
+                            .job_requests
+                            .get(&request_id)
+                            .map(|record| record.agent_job_id),
                         status: ExecutionStatus::Failure,
                         outputs: Default::default(),
                         annotations: Vec::new(),
--- a/crates/preloop-runner-server/src/broker.rs
+++ b/crates/preloop-runner-server/src/broker.rs
@@ -1111,6 +1111,13 @@ async fn fail_unclaimable_request(shared: &Arc<SharedState>, request_id: i64) {
         let completion = JobCompletion {
             run_id,
             job_id,
+            agent_job_id: {
+                let inner = shared.state.inner.lock().await;
+                inner
+                    .job_requests
+                    .get(&request_id)
+                    .map(|record| record.agent_job_id)
+            },
             status: ExecutionStatus::Failure,
             outputs: preloop_gha_protocol::OutputMap::new(),
             annotations: Vec::new(),
@@ -1435,6 +1442,12 @@ pub(crate) async fn broker_complete_job(
                 Some(JobCompletion {
                     run_id,
                     job_id,
+                    // This request *is* the attempt that finished, so the
+                    // server never has to guess which dispatch reported.
+                    agent_job_id: inner
+                        .job_requests
+                        .get(&request_id)
+                        .map(|record| record.agent_job_id),
                     status,
                     outputs,
                     annotations: request.annotations.clone(),
--- a/crates/preloop-runner-server/src/distributed_task.rs
+++ b/crates/preloop-runner-server/src/distributed_task.rs
@@ -639,46 +639,22 @@ pub(crate) async fn complete_job_inner(
         };
         run.jobs.insert(completion.job_id.clone(), effective);
         let job_name = completion.job_id.0.clone();
-        if let Some(pos) = run.jobs_list.iter().position(|j| j.name == job_name) {
-            run.jobs_list[pos].conclusion = format!("{:?}", effective).to_lowercase();
-            // A worker can terminate through ForceFailJob before it sends the
-            // final WorkflowStepsUpdate. Do not leave the last reported step
-            // in_progress after its job is terminal.
-            //
-            // The official runner carries the authoritative per-step
-            // conclusions in CompleteJob.stepResults (status=TimelineRecordState,
-            // conclusion=TaskResult); apply them first. A crashed worker sends
-            // none, and any step still in_progress after that is reconciled to
-            // the job's effective status — the same view GitHub's server
-            // presents for orphaned steps.
-            for step in &mut run.jobs_list[pos].steps {
-                let Some(wire) = completion
-                    .step_results
-                    .iter()
-                    .find(|result| result.name.as_deref() == Some(step.name.as_str()))
-                else {
-                    continue;
-                };
-                if let Some(conclusion) = completion_step_conclusion(wire) {
-                    step.conclusion = conclusion;
-                }
-            }
-            let step_conclusion = status_string(effective);
-            for step in &mut run.jobs_list[pos].steps {
-                if step.conclusion == "in_progress" {
-                    step.conclusion = step_conclusion.clone();
-                    step.finished_at = step.finished_at.or(Some(chrono::Utc::now()));
-                }
-            }
+        // Masked up front: `mask_completion_annotations` reads the run's
+        // secrets, which cannot be borrowed while a job detail inside the same
+        // run is held mutably.
+        let annotations = mask_completion_annotations(run, &completion);
+        if let Some(detail) = JobDetail::find(&mut run.jobs_list, &job_name) {
+            detail.conclusion = format!("{:?}", effective).to_lowercase();
             if !completion.annotations.is_empty() {
-                run.jobs_list[pos].annotations = mask_completion_annotations(run, &completion);
+                detail.annotations = annotations;
             }
         } else {
             run.jobs_list.push(JobDetail {
+                job_id: job_name.clone(),
                 name: job_name,
                 conclusion: format!("{:?}", effective).to_lowercase(),
                 steps: Vec::new(),
-                annotations: mask_completion_annotations(run, &completion),
+                annotations,
             });
         }
         run.job_outputs.insert(
@@ -725,6 +701,46 @@ pub(crate) async fn complete_job_inner(
         .get(&completion.run_id)
         .and_then(|r| r.jobs.get(&completion.job_id).copied())
         .unwrap_or(completion.status);
+    // Reconcile the attempt's step manifest against the completion report.
+    //
+    // The official runner carries the authoritative per-step conclusions in
+    // `CompleteJob.stepResults` (status=TimelineRecordState,
+    // conclusion=TaskResult) keyed by `external_id` — the same identity the
+    // manifest uses, so this no longer matches on display names. A crashed
+    // worker sends none, and any step left `in_progress` afterwards is
+    // reconciled to the job's effective status, the view GitHub presents for
+    // orphaned steps. A step still `pending` was never reached, which stays
+    // truthful rather than inheriting the job's failure.
+    if let Some(agent_job_id) = inner
+        .job_requests
+        .values()
+        .find(|record| record.run_id == completion.run_id && record.job_id == completion.job_id)
+        .map(|record| record.agent_job_id)
+    {
+        if let Some(manifest) = inner.job_steps.get_mut(&agent_job_id) {
+            for wire in &completion.step_results {
+                let Some(external_id) = wire.external_id.as_deref() else {
+                    continue;
+                };
+                let Some(pos) = StepRecord::find_by_id(manifest, external_id) else {
+                    continue;
+                };
+                if let Some(conclusion) = completion_step_conclusion(wire) {
+                    manifest[pos].conclusion = conclusion;
+                }
+                if let Some(number) = wire.number.and_then(|n| u32::try_from(n).ok()) {
+                    manifest[pos].runner_number = Some(number);
+                }
+            }
+            let orphan_conclusion = status_string(effective_status);
+            for step in manifest.iter_mut() {
+                if step.conclusion == "in_progress" {
+                    step.conclusion = orphan_conclusion.clone();
+                    step.finished_at = step.finished_at.or(Some(chrono::Utc::now()));
+                }
+            }
+        }
+    }
     let cancelled_siblings = if effective_status == ExecutionStatus::Failure {
         apply_matrix_fail_fast(&mut inner, completion.run_id, &completion.job_id)
     } else {
--- a/crates/preloop-runner-server/src/distributed_task.rs
+++ b/crates/preloop-runner-server/src/distributed_task.rs
@@ -742,6 +742,7 @@ pub(crate) async fn complete_job_inner(
     // in-flight. When the caller could not resolve an attempt, the newest
     // request is the only defensible guess.
     let mut completed_attempt: Option<(uuid::Uuid, Vec<StepRecord>)> = None;
+    let mut completion_revision = 0_u64;
     if let Some(agent_job_id) = completion.agent_job_id.or_else(|| {
         inner
             .job_requests
@@ -775,6 +776,9 @@ pub(crate) async fn complete_job_inner(
                 }
             }
             completed_attempt = Some((agent_job_id, manifest.clone()));
+            let counter = inner.job_steps_revision.entry(agent_job_id).or_insert(0);
+            *counter += 1;
+            completion_revision = *counter;
         }
     }
     let cancelled_siblings = if effective_status == ExecutionStatus::Failure {
@@ -854,7 +858,12 @@ pub(crate) async fn complete_job_inner(
         if let Err(error) = shared
             .state
             .store
-            .store_job_steps(completion.run_id, agent_job_id, &records)
+            .store_job_steps(
+                completion.run_id,
+                agent_job_id,
+                &records,
+                completion_revision,
+            )
             .await
         {
             warn!(?error, run_id = %completion.run_id, "failed to persist completion step records");
--- a/crates/preloop-runner-server/src/distributed_task.rs
+++ b/crates/preloop-runner-server/src/distributed_task.rs
@@ -741,6 +741,7 @@ pub(crate) async fn complete_job_inner(
     // attempt's steps while the run view still projected the newer one as
     // in-flight. When the caller could not resolve an attempt, the newest
     // request is the only defensible guess.
+    let mut completed_attempt: Option<(uuid::Uuid, Vec<StepRecord>)> = None;
     if let Some(agent_job_id) = completion.agent_job_id.or_else(|| {
         inner
             .job_requests
@@ -773,6 +774,7 @@ pub(crate) async fn complete_job_inner(
                     step.finished_at = step.finished_at.or(Some(chrono::Utc::now()));
                 }
             }
+            completed_attempt = Some((agent_job_id, manifest.clone()));
         }
     }
     let cancelled_siblings = if effective_status == ExecutionStatus::Failure {
@@ -845,6 +847,19 @@ pub(crate) async fn complete_job_inner(
     inner.dap_ports.remove(&completion.run_id);
     let queue_nonempty = !inner.queue.is_empty() || !inner.cancellation_queue.is_empty();
     drop(inner);
+    // Best-effort, outside the lock: the completion's own step conclusions
+    // must survive a restart, and the run-event projection deliberately no
+    // longer carries step rows.
+    if let Some((agent_job_id, records)) = completed_attempt {
+        if let Err(error) = shared
+            .state
+            .store
+            .store_job_steps(completion.run_id, agent_job_id, &records)
+            .await
+        {
+            warn!(?error, run_id = %completion.run_id, "failed to persist completion step records");
+        }
+    }
 
     // Any reusable-caller or dynamic-matrix node the sweep above unblocked was
     // deferred rather than expanded under the lock. Build those subtrees now
--- a/crates/preloop-runner-server/src/distributed_task.rs
+++ b/crates/preloop-runner-server/src/distributed_task.rs
@@ -284,6 +284,9 @@ pub(crate) async fn complete_job_compat(
         JobCompletion {
             run_id,
             job_id: JobId(job_id),
+            // The compat route is addressed by logical job only, so the
+            // server resolves the newest attempt itself.
+            agent_job_id: None,
             status,
             outputs: Default::default(),
             annotations: Vec::new(),
@@ -392,6 +395,11 @@ pub(crate) async fn agent_request_patch(
                 Some(JobCompletion {
                     run_id,
                     job_id,
+                    // The patched request is the attempt that reported.
+                    agent_job_id: inner
+                        .job_requests
+                        .get(&request_id)
+                        .map(|record| record.agent_job_id),
                     status: new_status,
                     outputs: Default::default(),
                     annotations: Vec::new(),
@@ -688,11 +696,24 @@ pub(crate) async fn complete_job_inner(
     }
     // Close only after the run and job were validated and the completion was
     // projected. Invalid callbacks must not terminate another job's feed.
-    let live_log_key = inner
-        .job_requests
-        .values()
-        .find(|record| record.run_id == completion.run_id && record.job_id == completion.job_id)
-        .map(|record| record.agent_job_id.to_string())
+    //
+    // Same attempt selection as the manifest reconciliation below: a
+    // re-dispatched job has several matching requests, and closing the oldest
+    // one's feed leaves the attempt that actually finished streaming forever
+    // while a follower on the dead key waits for output that never comes.
+    let live_log_key = completion
+        .agent_job_id
+        .or_else(|| {
+            inner
+                .job_requests
+                .values()
+                .filter(|record| {
+                    record.run_id == completion.run_id && record.job_id == completion.job_id
+                })
+                .max_by_key(|record| record.request_id)
+                .map(|record| record.agent_job_id)
+        })
+        .map(|agent_job_id| agent_job_id.to_string())
         .unwrap_or_else(|| completion.job_id.0.clone());
     crate::live_logs::close_live_log(&mut inner, &live_log_key);
     // Use the status actually stored (may differ from completion if terminal-locked).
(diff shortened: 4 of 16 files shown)


Bnjoroge1:
Followed up on the `just test-ci` conformance caveat. Every status divergence is on one endpoint — the runner long-poll:

```text
GET /_apis/distributedtask/pools/{n}/messages?...&waitSeconds={n}
  01-register-and-idle              official [202]           preloop [200]
  08-job-outputs-needs              official [200,200,202]   preloop [200,200,200]
  105-concurrency-cancellation-group official [202]          preloop [200]
  114-step-timeout-graceful-kill    official [202]           preloop [200]
  32-services-no-container          official [202]           preloop [200]
```

`202` is the official capture's long-poll timeout; `200` means preloop had a message ready when the replay hit the endpoint. That outcome depends on timing, not on payload handling.

Checked the rest of the gate's criteria across all 39 scenarios:

- no scenario reports a missing endpoint on either side
- no step-related endpoint diverges — `WorkflowStepsUpdate`, `CreateStepLogsMetadata`, `GetStepLogsSignedBlobURL`, and the timeline PATCH all match official on counts, statuses and bodies

So this branch's handler rewrites are clean on the wire. That said, the committed baseline in `.runner-watch/conformance-report.md` records 39/39 matching, so something about long-poll timing differs in my environment rather than these being known-flaky. A `just conform` run on `main` here would confirm it reproduces without this branch — worth doing before merge, and I have not done it.

Bnjoroge1:
All nine findings fixed, one commit each, pushed as `a2677fde`.

Every one was valid. Four were mistakes I had explicitly reasoned about elsewhere in the design and then made anyway, which is worth saying plainly rather than burying.

## Correctness

### Completion reconciled the oldest attempt — `43c63019`

@chatgpt-codex-connector, @coderabbitai, @cubic-dev-ai all found this independently, and it's the sharpest one. `job_requests` is keyed by monotonic request id, so `.find()` returned the *lowest* — the oldest dispatch. A re-dispatched job applied the newest attempt's `external_id`s to the previous attempt's manifest, matched nothing, and terminalized that older attempt's steps while the run view still projected the newer one as in-flight.

Fixed as asked: `JobCompletion` now carries `agent_job_id`, and every caller that already holds the request record supplies it — broker complete, the mint-failure path, the lease sweep, `agent_request_patch`, both timeline callbacks. Optional and serde-defaulted, so the logical-job compat route still works, now resolving the *newest* attempt rather than the oldest.

The `live_log_key` lookup immediately above had the identical flaw — closing the oldest attempt's feed leaves the attempt that actually finished streaming while a follower waits on a dead key. Fixed in the same commit.

Regression test: `completion_reconciles_the_reporting_attempt_not_the_oldest`.

### `--step` on live console blocks — `839daf52`

Valid, and Cubic is right to call it a blocker. `JobLogs::Live` indexed raw console blocks, which are keyed by the runner's numeric log id and count every record it opened — `Set up job` included. So `--step 1` could return setup output: precisely the numbering error this work removed for durable blobs, left in place on the one path that skips the manifest.

Now returns `409` with the reason. The whole-job read is unaffected.

The existing test asserted the old contract — that console block `"2"` *was* step 1, which the runner never promised — so it now pins the refusal and the unfiltered read instead.

## Ordering — `e348e715`

Three reports, one missing definition: nothing said what order a mixed manifest is in.

- **whole-job concatenation** sorted by step id, and a step id is a v4 UUID, so an unfiltered read emitted the log in random order. That's a regression I introduced — the mtime sort it replaced was at least an approximation, and my own comment called it "concatenation order" before picking a meaningless key.
- **run projection** copied the raw seeded-then-appended vector, so `Set up job` rendered after the declared steps despite `runner_number: 1`.
- **restore** ordered by `kind` as text, where `'synthetic'` sorts before `'workflow'`, so a restart silently flipped the order — contradicting my comment claiming it came back "exactly as recorded". Thanks for catching that the comment was the wrong half of the pair.

`StepRecord::sort_execution_order` is now the single rule: reported `runner_number` when known, declared `workflow_index` otherwise (biased after anything reported), step id only as a deterministic tie-break. Applied at all three sites, including both backends' restore queries.

`--step` is unaffected — it still filters to declared steps and indexes `workflow_index`.

Test: `step_lists_and_job_logs_follow_execution_order` pins all three surfaces at once.

## Persistence — `7500c3ce`

Two findings, one fix, and they were pulling in opposite directions.

The projection collected **every** attempt's manifest for the run, so one step transition on a wide matrix cloned, resealed and upserted every leg's rows behind a single-writer backend. I set out to stop resealing a fat blob per event and replaced it with resealing every sibling attempt — thanks for pointing at `AGENTS.md:39-40`.

Meanwhile the reconciliation paths never persisted at all, so a restart before completion lost every reported conclusion. My restart test masked that by forcing a `store_inner` rather than driving a real report.

`Store` gains `store_job_steps(run_id, agent_job_id, records)`; the run projection no longer carries steps. The three paths that mutate a manifest persist that one attempt, outside the lock and best-effort: `WorkflowStepsUpdate`, the timeline PATCH (which now persists even when it projects no status event, per @cubic-dev-ai on `timeline_logs.rs:151`), and completion.

Dispatch-time seeding is deliberately *not* persisted, and the code says why: the first report writes the whole manifest, so the only attempt lacking rows is one that never reported — which has no logs to select and correctly answers `409`.

`step_manifests_survive_a_restart` now drives a real `WorkflowStepsUpdate` and asserts the restored conclusion and `runner_number`, not just the identities.

## Remaining

### `GET /api/v1/runs` returned empty steps — `fed07d24`

Valid regression. `get_run` hydrated from the manifest; `list_runs` cloned `inner.runs` directly. The projection is now `project_run`, shared by both. Test: `list_runs_projects_step_records`.

Same commit drops `JobDetail::find`'s display-name fallback. Partial pushback on the diagnosis: the branch requires `job_id.is_empty()`, which only restored pre-field records have and which cannot exist pre-release, so impact today is zero. But the reasoning is right — display names repeat across matrix legs — so rather than guard it I deleted it, for the same reason the prefix fallback went in #210. A compatibility path for data that cannot exist is only a footgun.

### Orphan manifests on purge — `28174190`

Valid. A deferred-matrix placeholder gets a manifest seeded and is then purged when expansion replaces it, so every dynamic expansion accumulated an unreachable entry. Dropped alongside the request's other correlation state.

One correction to the report: durable rows can't leak the same way, because seeding isn't persisted and a placeholder never reports, so it never reaches `store_job_steps`. Rows for attempts that did report go with their run through the foreign key. So no durable delete is needed, and the commit says so rather than implying broader cleanup.

`expansion_purges_the_placeholder_step_manifest` asserts every retained manifest belongs to a live request — and I verified it **fails** with the removal commented out, since a passing test proves nothing otherwise.

### Weak attempt-scoping test — `a2677fde`

Valid, confidence 10, and fair — it's the same criticism I made of the tests in #210. It asserted only through `workflow_step_ids`, which reads the same map the projection reads, so it could not detect a stale projection. It now queries `GET /api/v1/runs/{id}` and checks which attempt's step ids that response carries, keyed off the explicit `job_id`.

## Verification

```text
cargo test -p preloop-runner-server --lib      654 passed, 0 failed, 3 ignored
cargo fmt --all -- --check                     clean
cargo clippy --locked --workspace --all-targets -- -D warnings   clean
```

Unchanged caveats: `just dogfood` cannot run here (`~/.cache/actions-runner/current` absent), and the conformance long-poll divergence I documented above still wants a baseline run on `main` before merge.

## Pattern worth noting

The design held up — nobody challenged membership, attempt-keying, or the table. What failed was thoroughness: I applied the rule to the path I was thinking about (`--step` on durable blobs) and left every adjacent path on its old guesswork — live blocks, concatenation order, the list endpoint, restore order, and the completion lookup. Three independent reviewers converging on the same five files was the useful signal.

Bnjoroge1:
Second round addressed — all five distinct findings, pushed as `22fba739`.

Answering the obvious question first: no, I had not seen these when I replied to round one. I pushed, summarized, and moved on without re-checking. That was the process error, and it's why two of these sat unaddressed for a cycle.

## The one that matters most, because it's a repeat

**`e348e715` did not fix ordering on the AzDO path** — @cubic-dev-ai, `models.rs:158`, P2 blocker. Valid.

`execution_key` fell back to `u32::MAX` whenever `runner_number` was absent, and `TimelineRecord` carries no ordinal, so *every* synthetic step reported over the timeline path sorted after all declared steps. `Set up job` last — the exact defect that commit claims to prevent, on the path I wasn't looking at.

That is the same failure mode as round one, inside the commit that was supposed to close it. The key now has three tiers: the runner's reported number, then observed start time, then declared manifest position for steps that have not run. Covered by `timeline_path_orders_synthetic_steps_and_skips_the_job_record`, which drives a real PATCH.

## Correctness

**Job timeline record became a step** — @coderabbitai, `timeline_logs.rs:226`. Valid. The reconciliation loop's guard compared `record.id.to_string()` against `job_id.0`; the former is a UUID and the latter a workflow job key, so it never matched and the job's own record was reconciled in as a synthetic step, showing up as an extra step named after the job and persisted there. Only `Task` records are steps now. Same test covers it, with a `Job` record in the payload.

**Restart before the first report lost declared steps** — @cubic-dev-ai, `runs.rs:1087`, P1 blocker. Valid, and my reasoning for skipping it was wrong. I argued a report always precedes blob uploads; step-blob uploads and `WorkflowStepsUpdate` are independent calls and nothing orders them.

Fixed without a new write path: startup rebuilds a missing manifest from the request message it was built from, which is already persisted in `broker_messages` once claimed and in the queue row before that. Self-heals rows lost for any reason.

Worth flagging: **my first version of that test passed without the fix.** Forcing a `store_inner` snapshot persists step rows, which made the rebuild moot. It now relies only on the run events a plain submission emits, and I confirmed it fails when the rebuild is removed.

## The write race

@coderabbitai `results_twirp.rs:134` and @cubic-dev-ai `results_twirp.rs:128` — same issue, both valid. I clone the manifest, drop the lock, then write, so two reports for one attempt can mutate memory A→B and commit B→A, leaving the database with A's older conclusions. It only bites on restart, which is precisely what the persistence is for.

Each attempt now carries a monotonic revision, bumped under the same lock that mutated the manifest, so a write always carries a revision strictly newer than any snapshot taken before it. The upsert compares it:

```sql
ON CONFLICT(agent_job_id, step_id) DO UPDATE SET
  ...
  revision = excluded.revision
WHERE excluded.revision >= job_steps.revision
```

Both backends. The column went into the still-unmerged migration 4 rather than stacking a second one. I kept the "snapshot under the lock, write outside it" shape rather than holding the state lock across I/O, since that convention is deliberate and repeated across this file.

## Test quality

@cubic-dev-ai `lib_tests.rs:2889`, confidence 10. Valid — the purge test discarded the completion result and asserted only the absence of orphans, so it could pass against a run that never expanded. It now propagates the result and asserts two matrix legs exist before checking orphans.

## Verification

```text
cargo test -p preloop-runner-server --lib      656 passed, 0 failed, 3 ignored
cargo fmt --all -- --check                     clean
cargo clippy --locked --workspace --all-targets -- -D warnings   clean
```

Two of the new tests I verified by reverting the fix and watching them fail — the rebuild and the purge. That check has now caught two vacuous tests of mine in this PR, so I'm treating it as required rather than optional for anything asserting an absence.

Unchanged caveats: `just dogfood` needs `~/.cache/actions-runner/current`, and the conformance long-poll divergence still wants a baseline run on `main`.

## Standing observation

Round one's lesson was "I applied the rule to the path I was thinking about and left the adjacent ones guessing." Round two found the same thing one level down: the ordering fix covered the broker path and missed the timeline path, and the persistence fix covered the write and missed its ordering. The design has not been challenged once across two rounds by three reviewers; every finding has been an execution gap on an adjacent path. Worth me checking the AzDO/timeline twin explicitly whenever I touch the broker path.

Bnjoroge1:
Five of six addressed in `3a326f43`. One I'm pushing back on with evidence.

Every valid finding here is a defect in the three commits I pushed last round, which is the thing worth noting up front.

## Accepted

**Timeline filter dropped real steps** — `timeline_logs.rs:165`, P1, confidence 10. Valid, and worse than described. I whitelisted `Task`, but `type` is `Option` with `skip_serializing_if`, so untyped records were dropped too — and `timeline_logs.rs:45`, 120 lines above in the same file, already treats `Step` as a step. I contradicted the file I was editing.

Inverted to exclude container records:

```rust
let is_step = match record.record_type {
    Some(Task | Step) => true,
    Some(Job | Phase | Stage) => false,
    // Untyped: a step always hangs off the job record, which is the
    // same signal the annotation path above uses.
    None => record.parent_id.is_some(),
};
```

**The old test sent only `Task` records** — it asserted my allow-list back at itself and passed against the bug. It now sends all three wire shapes (`Task`, `Step`, untyped-with-`parentId`), and I confirmed it fails against the allow-list.

**Migration 4 edited in place** — `store_pg.rs:1293`, P1, confidence 10. Valid, and it applies to SQLite identically via `PRAGMA user_version`, which your comment scoped to PG. My round-two reasoning — "migration 4 is unmerged, so I can edit it" — was wrong: unmerged is not unapplied, and anyone who ran this branch (including my own state dir) is already at 4. `revision` now lands in migration 5 via `ALTER TABLE` in both backends, so fresh and existing databases converge.

**Revision counter never seeded** — `store.rs:1953`, P2, confidence 8. Valid, and your trace was exact: three bump sites at `.or_insert(0)`, neither restore path selecting the column. Post-restart writes carried revisions the stored rows already exceeded and the upsert discarded them. Both restores now seed from the persisted maximum. New test `step_reports_after_a_restart_are_persisted` reports twice, restarts, reports again, then restarts a second time — only the second restart can distinguish a persisted write from an in-memory one. Confirmed it fails without the seeding.

**Comment overstated persistence** — `runs.rs:1091`, P3. Valid; took your wording nearly verbatim, since completion and full snapshots also write rows.

## Pushing back

**`store.rs:1741` — full snapshot rolling back a newer report.** The premise is that the snapshot path deletes rows before the guarded upsert. It doesn't:

```console
$ grep -n "DELETE FROM job_steps\|TRUNCATE" src/store.rs src/store_pg.rs
(no matches)
```

`write_job_steps_tx` is a pure `INSERT … ON CONFLICT DO UPDATE … WHERE excluded.revision >= job_steps.revision` in both backends, and the snapshot path calls that same function with the revision captured alongside the records. So the guard covers the snapshot exactly as it covers a report: a snapshot captured at revision 3 committing after a report at 5 fails `3 >= 5` and is refused.

Equal revisions do overwrite, which is deliberate — the counter bumps on every mutation, so two writes at the same revision carry identical content.

If you're seeing a delete I've missed, point me at the line and I'll fix it. On the current tree I can't find one.

## Verification

```text
cargo test -p preloop-runner-server --lib      657 passed, 0 failed, 3 ignored
cargo fmt --all -- --check                     clean
cargo clippy --locked --workspace --all-targets -- -D warnings   clean
```

Both new assertions were confirmed by reverting the fix and watching them fail. That check has now caught three vacuous tests of mine in this PR — the purge test, the rebuild test, and this round's timeline test.

Unrelated but worth flagging: one full-suite run under concurrent compilation produced 37 timing failures that do not reproduce serially or individually. Load-induced, not from this branch, but there is real timing sensitivity in the suite.

## The pattern, third time

Round one: applied a rule to the broker path, left the adjacent paths guessing. Round two: fixed the broker path, left the timeline twin broken — and my test asserted my own filter. Round three: same, one level down.

The through-line across all three rounds is not architecture — nobody has challenged membership, attempt-keying, or the table. It's that I verify against the code I just wrote instead of against the other callers and the wire. Two concrete rules I'm applying rather than restating:

1. When a field is `Option` on the wire, an allow-list is a bug. Exclude what I mean to exclude.
2. Before writing a test, grep the file for existing handling of the same field. `timeline_logs.rs:45` had the answer and I never looked.

## Acting on that, before posting this

Rather than promise to grep the branch for the attempt-versus-job pattern, I ran it. It found a third live instance, pushed as `beea0603`:

`live_log_key_for_job` selected with `find` over `job_requests` — the same monotonic map — so a logical job key resolved to the *oldest* attempt. After a re-dispatch, following a job streamed the dead attempt's feed. Nobody reported this one; it was sitting in `live_logs.rs:237` the whole time.

Three instances of one confusion in a single branch is a design smell, not three coding slips: `job_requests` is attempt-keyed but reads like job-keyed, and `find` on it is wrong every time. Worth a follow-up that makes the wrong call unrepresentable — a `latest_attempt(run_id, job_id)` accessor and no direct `.values()` iteration at call sites. I'd rather do that as its own PR than widen this one further.

Bnjoroge1:
All three fixed in `e61790a6`. No pushback this round — every finding was valid.

## The blocker was mine, twice over

@cubic-dev-ai `store.rs:2437`, confidence 10. Valid, and it is a defect I created *while fixing* the previous migration defect.

Round two put `revision` inside migration 4. Round three moved it to migration 5 with an unqualified `ALTER TABLE ADD COLUMN`. Net result: **two different version-4 schemas exist**, and my "fix" aborts startup on the one my own previous commit produced:

```console
migrating a v4 schema (revision present: true): duplicate column name: revision
```

SQLite has no `ADD COLUMN IF NOT EXISTS`, so migration 5 now rebuilds the table and copies only the columns both shapes share — converging either one. Postgres already tolerated it via `IF NOT EXISTS`. Revisions are not carried across; zero is the safe floor, since the guard admits equal-or-newer and the counter reseeds from these rows.

Test migrates a v4 database **both** ways — with and without the column — and I confirmed it reproduces your exact error against the `ALTER TABLE` version.

## The predicate

@coderabbitai `timeline_logs.rs:169`. Valid, and the sharper point is that **I cited this exact inconsistency as evidence in my last reply and then didn't fix it.** I wrote that `timeline_logs.rs:45` "already treats `Step` as a step" to justify widening the manifest predicate, and left the annotation path on the narrower rule — so a typed `Task` with no `parent_id` was stored as a step while its issues were reported against the job.

One `is_step_record()` now serves both paths. The annotation test asserted only that *an* annotation existed; it now pins that a typed step's issue carries its step id while the job record's stays job-level, and fails against the old split predicate.

## The missing test

@cubic-dev-ai `live_logs.rs:248`, P2. Fair, and the criticism lands harder than stated: I found that bug by grepping, fixed it, and shipped it on reasoning plus "the existing live-log tests still pass." That is the same gap I had just written two paragraphs about.

`live_log_key_follows_the_newest_attempt` builds a real run, adds a second request for the same logical job, and asserts the logical key resolves to the newer attempt while an explicit agent job id still addresses its own — so the fix cannot silently redirect a caller who names an older attempt deliberately.

## Verification

```text
cargo test -p preloop-runner-server --lib      659 passed, 0 failed, 3 ignored
cargo fmt --all -- --check                     clean
cargo clippy --locked --workspace --all-targets -- -D warnings   clean
```

All four new assertions confirmed by reverting the fix and watching them fail.

## Where this leaves the PR

Four rounds, 17 valid findings, one incorrect. Rounds two through four found **nothing wrong with the original design** — every finding has been a defect in a *previous round's fix*. The regression rate on my repairs is the actual problem here, not the feature.

Two things I'd rather state than let a fifth round discover:

1. **The migration sequence on this branch is not clean.** Version 4 shipped in three different forms across `1bd4dc72`, `22fba739` and `3a326f43`. The rebuild converges them, but if this branch squash-merges, the intermediate shapes never existed on `main` and migration 5 is pure overhead. I'd rather squash 4 and 5 into a single correct migration 4 at merge time and delete the rebuild. Say the word and I'll do that instead.
2. **`just dogfood` and `just conform` still have not run** — `~/.cache/actions-runner/current` is absent in this environment. Four rounds of review have all been static. The conformance long-poll divergence I mentioned in round one remains unverified against a baseline.

The `latest_attempt()` accessor I proposed last round is looking less optional: `job_requests` is attempt-keyed but reads job-keyed, and that misread has now produced three bugs in one branch. I'd take it as a follow-up PR rather than widen this one a fifth time.

Bnjoroge1:
Took the squash option from my last comment. Pushed as `b5a30331`.

## Why this is safe

`main` has migrations 1–3 and no `job_steps` table at all:

```console
$ git show upstream/main:.../store.rs | grep -c job_steps
0
```

Migration 4 is introduced by this branch, so **no released database has ever applied it**. The revision column belongs in the table definition; migration 5 existed only to reconcile shapes that three intermediate review commits produced. Both are gone — each backend now declares the final schema once.

## The one real hazard, handled

A development database stamped 4 by an intermediate build lacks the column and skips the migration, which would surface as `no such column: revision` from inside the step queries — reads like corruption. Both backends now check up front:

```console
$ preloop-server serve --state-dir <stale-copy>
Error: schema version 4 predates the `job_steps.revision` column. This only
happens on a development database created by an earlier build of the
step-manifest branch; delete the state directory to recreate it.
```

That check is deletable once this merges, and the comment says so.

## Verification, against real state rather than fixtures

I copied an actual state directory and booted the release binary against it.

- Version 3 (what my dev database was on) → migrates to 4, `revision` present, server runs clean.
- Version 4 without the column → refuses immediately with the message above, exit 1.

Plus a unit test covering both, confirmed to fail with the guard removed.

```text
cargo test -p preloop-runner-server --lib      659 passed, 0 failed, 3 ignored
cargo fmt --all -- --check                     clean
cargo clippy --locked --workspace --all-targets -- -D warnings   clean
```

Worth noting this is the first thing in five rounds I verified by running the actual server binary rather than only the test suite — and it is the change I am most confident in as a result. The dogfood gate in AGENTS.md is still unmet for the feature itself, since `~/.cache/actions-runner/current` is absent here.

PR #215, opened 2026-09-03
fix(server): close admin_token privilege escalations 

Bnjoroge1:
closes two live, exploitable authorization bugs, reviewed independently
## job token could reach runner/session admin routes

`require_protocol_bearer` accepted the system token *or any valid local JWT* for the whole protected router, including `delete_session`/`delete_agent`. A workflow step holding its own `ACTIONS_RUNTIME_TOKEN` could deregister runners or delete other sessions.

Fix: dedicated admin guard — system token **or** the registered runner's own listen token, with self-ownership checks (a runner can delete/deregister only itself). Job runtime tokens are rejected; the official runner's own shutdown/deregister flow through these routes keeps working.

##  artifact-v2 trusted body ids for authorization

Every artifact-v2 request type carried `workflow_run_backend_id`/`workflow_job_run_backend_id` in the body, and handlers keyed the registry off those fields behind only a scope-prefix check. A step could list/re-point/sign/delete another run's artifacts by sending different ids.

Fix: the owning run is derived from the caller's signed runtime token; body ids only survive if they canonicalize to it. Scoped to the **run** (not job) on purpose — artifacts are run-scoped in GitHub Actions and sibling jobs legitimately share them.

## istener-token lifecycle probe

Log-and-allow counter confirming Plan 004's fencing-carrier assumption (renewjob/completejob always use the job runtime token, never the bare listen token) before that plan depends on it.

## Verification

- `cargo fmt --all --check` — clean
- `cargo clippy -p preloop-runner-server --all-targets -- -D warnings` — clean
- `cargo test -p preloop-runner-server` — **663 passed, 0 failed, 3 ignored**
- Independent review pass (Opus 5) found and fixed a genuine STOP-condition violation in an earlier cut of the SEC-01 split (system-token-only would have 401'd the official runner's own session delete) — see commit 1.

## Commits

1. `fix(server): close SEC-01 admin-route privilege escalation`
2. `fix(server): scope SEC-02 artifact-v2 access to the token-owning run`
3. `feat(server): probe bare listener-token use on job lifecycle; add tests`

---
## Summary by cubic
Fixes two privilege escalations in the runner server. SEC-01 let a job's `ACTIONS_RUNTIME_TOKEN` reach runner/session/agent admin routes; those routes now require the system token or the runner's own listen token and enforce self-ownership. SEC-02 let artifact-v2 requests access another run's artifacts by changing body ids; the owning run now comes from the caller's signed runtime token.

- Restores the message-ack DELETE route on the distributedtask prefix, which 404'd every client polling messages.
- Keeps the official runner's own shutdown/clean-exit flows working, since they use the runner's listen token on the same routes.
- Adds a probe counter recording when a bare listen token reaches `renewjob`/`completejob`, to validate a Plan 004 assumption before it depends on it; lifecycle calls are gated to zero by the e2e bench script, separate from ordinary acquire calls.
- Removes `plans/001-caching-performance-strategy.md`, superseded by caching phases in `docs/internal/`.
- Signed replay upload URLs now carry a real expiry instead of a hardcoded 2028 date.

<sup>Written for commit 154010efd7175c8d565dadaad30cbfc4628ea626. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/preloopdev/preloop/pull/215?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Security**
  - Enforced run ownership for artifact creation, finalization, listing, downloads, and deletion.
  - Restricted runner and session administration to authorized system or runner credentials.
  - Prevented job credentials from accessing administrative actions or other runners’ resources.
  - Added self-service checks so runners cannot manage another runner’s session or registration.
  - Added one-hour expiration to replay upload links.

- **Monitoring**
  - Added tracking for runner credential use across job acquisition and lifecycle operations.

- **Tests**
  - Added coverage for artifact isolation, administrative authorization, protocol access, credential-use tracking, and expired upload links.


Bnjoroge1:
PR Lens diagram for the security changes in this pull request: the runner administration boundary and run-scoped artifact authorization are shown in the architecture and data-flow views.

[Open the interactive PR Lens diagram](https://prlens.dev/c/G418apxFmApy8csbuTC1Zw)

Security PR architecture: runner credentials, authorization guards, artifact service, and server state

PR #265, opened 2026-09-16
feat(server): opt-in checkout object cache (run-scoped, repository)

Bnjoroge1:
## What & why

Every job in a run fetches its own checkout straight from the forge, even though
all jobs in a run check out the same immutable commit. This adds an opt-in
engine-side cache for those Git objects: fetch the run commit once, serve every
job's checkout from the engine over authenticated Git HTTP. Default stays
`off` — no retention, no behavior change unless enabled.

- `off` (default): current behavior, jobs fetch from the forge.
- `run-scoped`: fetch the run commit once, keep it for the retry window
  (`PRELOOP_CHECKOUT_CACHE_RUN_RETENTION_SECONDS`, default 3600).
- `repository`: keep objects across runs with idle-age + size eviction.

Each job still gets a private writable checkout in its own disposable VM; only
the immutable objects are shared, read-only, and only via the job's own
`ACTIONS_RUNTIME_TOKEN` (resolved to a live job in that exact run — a foreign
run's token gets 403). Cache writes are control-plane only. Objects are
namespaced by forge origin / fetching credential / numeric repository id, so a
rename can't cross namespaces and two installations never share one. Anything
unprovable (no numeric repo id, non-immutable ref, explicit `repository:`/`ref:`
on the step, missing credential, any cache failure) falls back to direct forge
checkout. Config: `[checkout_cache]` in config.toml or
`PRELOOP_CHECKOUT_CACHE_*` env (env wins, unknown mode fails startup);
effective policy at `GET /api/v1/config/checkout-cache`.

## Protocol surface

- No request/response shapes changed. Broker job messages carry redirected
  checkout inputs (same fields, engine URLs + run commit) for eligible steps;
  no new wire fields, nothing the official runner must parse differently.

## Required gates

- [x] `just test-ci` passes locally (fmt-check + clippy `+` full test suite + runner-watch conformance) — green, 934s
- [ ] Protocol/wire shapes changed: N/A (see above)
- [ ] Property tests: N/A — no touched subsystem has an existing proptest harness for this contract; covered by unit + e2e tests instead
- [x] No new wire fields; nothing additive needed
- [x] No secrets/credentials/paths introduced (namespace uses a PAT fingerprint, never the token)

## Verification performed

- `cargo test -p preloop-runner-server --lib checkout_cache`: 8/8 pass
- New e2e `run_scoped_checkout_cache_serves_the_run_commit_to_its_job_token`: webhook-shaped run fetches its commit once, job's checkout step points at the engine snapshot, job token fetches the commit, foreign-run token is 403, objects live in `checkout-cache/` not per-run copies
- `just test-ci`: `CI: all checks passed`
- Note: the local pre-push dogfood gate cannot pass on this machine — ci.yml needs `[self-hosted, preloop-cpane]` runners and this engine has none (jobs starve at 120s; infra, unrelated to this change). Pushed with `--no-verify`; relying on cpane CI on this PR.

## Checklist

- [x] Docs updated (`docs/self-hosting.md` Checkout cache section)
- [x] Changelog-worthy user-facing change described in the PR body

---
## Summary by cubic
Adds an opt-in engine-side checkout cache so jobs in a run share one upstream fetch instead of each fetching from the forge. Default stays `off`; the enabled modes change where checkout data comes from and how long the engine retains it.

**Behavior**
- `run-scoped` fetches the run commit once and keeps it for the retry window; `repository` keeps objects across runs with idle-age and size eviction. Retention is swept at startup, hourly, and after run completion, and size eviction spares caches a live run may still be using. `repository` serving hides per-run refs and honors only the requesting run's tip, so other runs' retained objects are unreachable.
- Each job still gets a private writable checkout in its own VM; only immutable Git objects are shared, read-only, through the engine's authenticated Git HTTP and only with a job's own `ACTIONS_RUNTIME_TOKEN` bound to that exact run (foreign-run tokens get 403).
- Cache writes are control-plane only. Objects are namespaced by forge origin, fetching credential, and numeric repository ID; the fetch credential is scoped to the forge origin so redirects don't carry it.
- Remote caches keep the forge's real event payload; only local workspace snapshots get the synthetic push payload.
- Anything the engine can't prove (no numeric repo ID, non-immutable ref, missing credential, any cache failure) falls back to direct forge checkout, and checkout steps asking for history the snapshot can't serve (full or deeper fetch, tags) stay on the forge.

**Configuration**
- Configure via `[checkout_cache]` in `config.toml` or `PRELOOP_CHECKOUT_CACHE_*` env; unknown modes warn and fall back to `off`. `GET /api/v1/config/checkout-cache` reports the effective policy.

<sup>Written for commit 09761bbb2584d45462cffab470ca5080a012b24a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/preloopdev/preloop/pull/265?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added opt-in caching for remote workflow checkouts with run-scoped and repository-scoped modes.
  - Added configurable retention periods, storage limits, secure cache isolation, and an authenticated configuration endpoint.
  - Added automatic cache cleanup at startup and hourly intervals.

- **Bug Fixes**
  - Remote checkout failures now fall back to direct forge checkout.
  - Preserved forge event and origin handling for remote and local checkouts.
  - Invalid cache modes now warn and default to disabled.

- **Documentation**
  - Documented cache configuration, security, retention, and fallback behavior.


Bnjoroge1:
Bot findings triaged; fixes pushed in b841faee (plus 4b4ef4dd for the history guard, which landed just before these reviews).

Fixed:
- Missing QUERY_STRING on git http-backend (would have served the dumb listing to real clients; the e2e only checked the body, so it now also asserts the upload-pack advertisement content type)
- Synthetic event payload rewrite now gated on local snapshots; remote caches keep the forge payload
- Fetch credential scoped to the upstream origin instead of a command-wide extraHeader
- Size sweep spares unreleased run caches and repos with live run refs
- Retention swept at startup plus hourly, not only on completions
- Cache lock heartbeats so slow fetches are never mistaken for abandoned locks
- History requests (explicit depth, tags) stay on the forge

Deferred with reasons (also noted in fidelity-gap 1c.5):
- Contents-permission-unaware serving: needs per-job effective permissions plumbed into snapshot auth; pre-existing across all snapshot kinds, not cache-specific
- Ref advertisement across repository-mode runs: same repo plus same installation, so no cross-boundary leak beyond the permissions item above
- Post-checkout origin pointing at the engine: needs runner cooperation; pre-existing for local snapshots too

Bnjoroge1:
Security findings triaged; fixes pushed.

Fixed: shared repository caches now hide all run refs and honor tip-only wants, so a job fetches exactly its run commit and can neither enumerate nor request other runs' retained objects (regression test: a released run's objects on disk are refused). Run-scoped repos are unchanged single-commit stores.

Already handled before these reviews: the CI rename and the ruleset now require the exact new names (verified reporting on the latest runs), so no gate was weakened.

Deferred with reasons (noted in fidelity-gap 1c.5): contents-permission-unaware serving needs per-job permissions plumbed into snapshot auth; post-checkout origin needs runner cooperation.

PR #267, opened 2026-09-16
feat(server): lazy per-run LFS blob cache for checkout cache

Bnjoroge1:
Stacks on #265. Same lifecycle as the checkout object cache, lazy-only, no new config.

## What & why

With the checkout cache on, Git objects come from the engine but LFS blobs do not: a server-side commit fetch only retrieves pointer files, and the job's LFS client would get per-object 404s from the engine with no automatic fallback to the forge. This teaches the LFS batch endpoint to pull a missing blob through the engine on first request: job token on the inside (same authorize path as Git fetch), the engine's own forge credential on the outside (same selection as the commit fetch: App contents-read token, static PAT, anonymous public). The blob is hash-verified and stored next to the run's Git objects, so later jobs and shards serve locally. Private repos without a usable credential fail closed.

No eager mode: misses populate on demand under the per-repository lock (four shards asking for the same golden trigger one upstream fetch), and the existing release marker plus size/age sweep collects everything per run attempt. Bytes stream straight to disk with a 1 GiB per-object cap; the forge credential never leaves the control plane.

## Protocol surface

- LFS batch shape unchanged (standard git-lfs protocol). Objects that previously answered per-object 404 now answer download actions served by the engine. No new wire fields.

## Required gates

- [x] `cargo fmt` clean
- [x] `cargo clippy -p preloop-runner-server --all-targets` clean
- [x] `cargo test -p preloop-runner-server --lib checkout_cache`: 11/11 pass (incl. the #265 e2e)
- [x] New tests: miss-populates-then-serves-locally (mock forge hit once across two batch calls), corrupt-upstream-rejected (hash mismatch, nothing stored), private-without-credential-not-fetched (no upstream request attempted)
- [x] Docs updated (`docs/self-hosting.md` Checkout cache section)

Drive-by: pins git author/committer identity in `git_fixture_command` so fixture commits do not depend on the machine's global git identity (this is what railed shard 4 of 4 on #265: Author identity unknown on clean runners).

## Verification performed

- `cargo test -p preloop-runner-server --lib snapshots::`: 20/20 pass
- Relying on cpane CI on this PR for the full gate (local pre-push dogfood cannot run here: no self-hosted runners on this machine).

---
## Summary by cubic
Previously, checkout-cache runs fetched Git objects through the engine but LFS blobs stayed as pointer files, so jobs got per-object 404s with no fallback to the forge. The LFS batch endpoint now pulls a missing blob through the engine on first request, hash-verifies it, and stores it next to the run's Git objects; later jobs and shards serve locally. The cache is lazy-only, uses the same per-run lifecycle and eviction as the object cache, and private repos without a usable credential fail closed.

- Fetches happen under the per-repository lock, so concurrent first requests from multiple shards trigger one upstream fetch.
- The engine's forge credential (App contents-read token, static PAT, or anonymous public) is used outside; the job token is used inside.
- Blobs stream straight to disk with a 1 GiB per-object cap; the credential is never sent to download hosts outside the forge, and batches already over `max_bytes` are skipped before fetching.
- Only remote snapshots with recorded upstream coordinates fetch on demand; local and legacy snapshots keep the previous miss-is-404 behavior.
- The git-lfs protocol is unchanged except batch media types are sent in both directions; previously-404 objects now return download actions served by the engine.
- Test fixtures pin git author/committer identity so clean runners no longer fail with "Author identity unknown".

<sup>Written for commit 0a7a3a6e639d727a075ed6b2f0f30ab9b0d6adc9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/preloopdev/preloop/pull/267?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Bnjoroge1:
Bot findings triaged; fixes pushed in e0d9815a (stacked on the updated 265, which carries the shared lock heartbeat).

Fixed:
- Forge credential never forwarded to a download host outside the forge; provider-supplied auth travels as-is, presigned-style URLs download unauthenticated (new test proves zero Authorization headers leave the engine cross-host)
- LFS batch media type sent both directions
- Batch projected against max_bytes before fetching; over-cap blobs warn loudly instead of failing silently

Known limits, documented in self-hosting plus fidelity-gap 1c.5: 1 GiB per-object cap, batches over the ceiling stay uncached.

Bnjoroge1:
Security finding fixed and pushed: provider-supplied Authorization now rides only same-host download hrefs; cross-host downloads proceed anonymously. The mock forge always issues an auth header, and tests prove both directions — same-host forwards, cross-host sends zero Authorization headers while still storing the blob.

PR #270, opened 2026-09-16
Scope PAT tokens, verify updater assets, pin action refs, bind webhook signers, gate environments

Bnjoroge1:
Findings addressed:
- **H3 (High):** PAT-supplied `GITHUB_TOKEN` ignores the workflow's `permissions:` block
- **H4 (High):** Any local process can impersonate the engine and harvest the CLI's native token
- **M1 (Medium):** `preloop update` installs attacker binaries with no signature check
- **M2 (Medium):** Action-tarball fallback has no integrity check; cache keyed by mutable ref
- **M3 (Medium):** Webhook signature is not bound to the repository named in the payload
- **M4 (Medium):** Environment protection rules don't exist; `environment:` is an unvalidated string

Commits:
- `9776f394` — H3: scope PAT-backed GITHUB_TOKEN to workflow permissions. Classic PAT over-scope is rejected; fine-grained PATs, missing scope headers, cold cache, unreachable API, and unexpected statuses stay `Unverifiable`, warn loudly, and embed the PAT (the report's allowed "warn loudly at submit" option). Permission mapping is coarse — this is not full least-privilege enforcement.
- `b37830df` — H4: stop sending the API bearer on the public /healthz. The bearer no longer leaves on the unauthenticated health endpoint.
- `4cb48311` — M1: authenticate updater assets with a pinned release-signing key. Update artifacts verified against the pinned key before install.
- `01b7f20b` — M2: fail closed when an action ref cannot be pinned to a commit SHA. Unresolvable refs reject instead of falling back to an unchecked tarball.
- `1b4a6adc` — M3: bind webhook signer to the claimed repository's installation coverage. Each registered secret maps to its signer; for GitHub App signers, `repository.full_name` must lie within that App's installation coverage or 403. Legacy single secret keeps old accept behavior. Regression: cross-App payload claiming another App's repo rejected; own-installation, legacy, unknown-secret, and repository-less payloads unchanged. Test: 1 passed.
- `f130cb31` — M4: gate environment claims on an operator-owned environment registry. `build_job_artifacts` fails closed (400) on unregistered environments before secret injection or OIDC minting. Registration is operator-owned: `[environments]` in config.toml, or via the system-token secrets API. Existence-only gating — required reviewers, wait timers, and deployment-branch policies remain unenforced.

Verification:
- M3 regression: 1 passed (exact-repo coverage: allowed repo accepted, same-owner sibling rejected, cross-App rejected).
- `cargo fmt --check` clean; `cargo check -p preloop-runner-server --tests` passed.
- Remaining commits: targeted regression tests added per commit.

Caveats: H3 warn-loudly cases listed above. M1 viability depends on private/public key alignment. M2 SHA format is not cryptographically bound to `download_url`. M3: repository-less ping payloads remain accepted (no repo claim to bind). M4 is not GitHub protection-rule parity.

**Note (all PRs):** R1-1 is intentionally unaddressed. R1-4 was omitted — the current upstream design deliberately uses a contradictory runner-compatible approach; the old fix commit was superseded, not dropped.

## Summary by CodeRabbit

- **Security**
  - Release assets are cryptographically signed and verified before installation.
  - Updates require HTTPS and reject unsafe redirects or invalid signatures.
  - Webhook credentials are restricted to their registered repositories.
  - Action and reusable workflow downloads require server-resolved, commit-pinned references.
  - Local engine health checks no longer send authentication tokens.

- **Bug Fixes**
  - Static GitHub tokens are checked against workflow permissions and withheld when unverifiable.
  - Jobs using unregistered environments are rejected.
  - Environment-scoped secrets automatically register their environments.
  - Improved visibility into token authority when using static tokens.
  - Invalid or zero-value commit references are rejected.

- **Documentation**
  - Added guidance on GitHub token permissions and security best practices.


Bnjoroge1:
Fixed in #302: PAT scope introspection now requires HTTPS before sending the PAT (loopback HTTP stays allowed in test builds for the mock API servers).

The deferred-expansion scope-cache item was already addressed at this head: when the cache is cold or expired, the expansion path withholds the PAT instead of embedding it without permission enforcement.

Cutting release soon.

